PR #18, opened 2014-12-15
DNET-570 Loading error messages from a manifest resource fails on mono

mrgleba:
Added a call to ResourceSet.GetEnumerator to initializa the ResourceSet
Added caching of ResourceSets

### DIFF
--- a/NETProvider/src/FirebirdSql.Data.FirebirdClient/Common/IscException.cs
+++ b/NETProvider/src/FirebirdSql.Data.FirebirdClient/Common/IscException.cs
@@ -39,9 +39,13 @@ namespace FirebirdSql.Data.Common
 		private string _message;
 		#endregion
 
-		#region · Properties ·
+        #region · Static Fields ·
+        private static Dictionary<string, ResourceSet> _resources = new Dictionary<string, ResourceSet>();
+        #endregion
 
-		public List<IscError> Errors { get; private set; }
+        #region · Properties ·
+
+        public List<IscError> Errors { get; private set; }
 
 		public override string Message
 		{
@@ -194,10 +198,8 @@ namespace FirebirdSql.Data.Common
 			// step #2, see if we can find a mapping.
 			else
 			{
-				using (var rs = CreateResourceSet("FirebirdSql.Data.Resources.sqlstate_mapping"))
-				{
-					this.SQLSTATE = rs.GetString(this.ErrorCode.ToString());
-				}
+				var rs = CreateResourceSet("FirebirdSql.Data.Resources.sqlstate_mapping");
+				this.SQLSTATE = rs.GetString(this.ErrorCode.ToString());
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


cincuranet:
HI, do you have any measurements about caching the `ResourceSet`? I like the idea, although I don't want to have unnecessary objects in memory for long time for not real benefit.

mrgleba:
Hi!

My reasoning is that memory is pretty abundant and IO is expensive.
There are about 1000 records in each: is_error_msg and sql_state_mapping.
It seems the cache would take circa 100kb in mem, which IMHO is acceptable.

Alternatively the cache could be encapsulated in a WeakReference soas to
get GCed when mem becomes scarce.

Cheers!
Michał




2015-01-13 8:53 GMT+01:00 Jiri {x2} Cincura <notifications@github.com>:

cincuranet:
Just thinking... Maybe it's worth to just create two static hashtables (one for each resource) and remove the resources whatsoever. It will be in the code. Immediately available. No need to fiddle with resources.

That might be even better, don't you think?

mrgleba:
Makes complete sense for sql_state_mapping.
As for error messages I wouldn't change it in order to enable localizations
in the future

Cheers!
Michał

2015-01-13 14:34 GMT+01:00 Jiri {x2} Cincura <notifications@github.com>:

cincuranet:
Good point. But in last 5+ years nobody did anything about localization of messages. And `firebird.msg` is kept in English anyway. 

I'll give it a day to think about and we'll see.

mrgleba:
>
> Good point. But in last 5+ years nobody did anything about localization of
> messages. And firebird.msg is kept in English anyway.
>
Still it'd be good to have the possibility. The cost of loading it to the
cache from resources once is negligeble.

>
Cheers

cincuranet:
Hi. So I made some changes. I decided to make it all static. When the translation comes and somebody will be willing to keep it updated (I highly doubt that's going to happen), I might change it.

Can you have a look at [this branch](https://github.com/cincuranet/NETProvider/tree/resources)? If it looks good, I'll merge it and release.

[Here](https://ci.appveyor.com/project/cincura_net/netprovider/build/0.0.0.141)'s the - curently latest - build of that branch.

mrgleba:
Hi!

> Can you have a look at this branch
> <https://github.com/cincuranet/NETProvider/tree/resources>? If it looks
> good, I'll merge it and release.
>
I'd use TryGet in accesing SqlStateMapping. Apart from that the change
seems fine.

Cheers!
Michal

cincuranet:
Why?

mrgleba:
We'd get an exception in there in case we recieve an unknown error code.

2015-01-21 8:53 GMT+01:00 Jiri {x2} Cincura <notifications@github.com>:

cincuranet:
No. You would not. See [this](https://github.com/cincuranet/NETProvider/blob/resources/NETProvider/src/FirebirdSql.Data.FirebirdClient/Common/IscException.cs#L197). It's calling the [`GetValueOrDefault`](https://github.com/cincuranet/NETProvider/blob/resources/NETProvider/src/FirebirdSql.Data.FirebirdClient/Common/IscException.cs#L284) helper.

mrgleba:
You're right.
I was looking at your initial commit rather than the final result.
That's why you shouldn't answer e-mails before your morning coffee :)

Everything looks fine.

Cheers
Michał

2015-01-21 9:02 GMT+01:00 Jiri {x2} Cincura <notifications@github.com>:

cincuranet:
Good, I'll run the test suite and merge.

cincuranet:
Reworked into PR #22.

mrgleba:
Great!
Thank you.

2015-01-23 8:13 GMT+01:00 Jiri {x2} Cincura <notifications@github.com>:

cincuranet:
I'll try to release new version in few days.

PR #19, opened 2014-12-24
DNET-582

zabulus:
Random bug in finalizing DbCommands

### DIFF
--- a/NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbCommand.cs
+++ b/NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbCommand.cs
@@ -403,6 +403,9 @@ namespace FirebirdSql.Data.FirebirdClient
 
 						this.disposed = true;
 					}
+					catch
+					{
+					}
 					finally
 					{
 						base.Dispose(disposing);


zabulus:
@cincuranet will you merge this PR? I know you don't prefer such band aid solution. But I can see that there is the same crutch exists in ```GdsStatement.Dispose```

cincuranet:
This is hard. It's such a hack.

I'd rather have a test app that fails once an hour. Then I'll iron the bug out no matter what.

zabulus:
In our case it is a server with a lot clients, that randomly crashes with OOM. Exception halts finalizer thread and all objects with finalizers begin overrun finalizer queue.

cincuranet:
Can you try to create something that crashes under a simulated load or something like that? It doesn't need to be 100%. If it crashes once an hour, it's a good start.

OT: Interesting the finalizer thread will not recover.

zabulus:
I've found this bug in our CI nightly build. 99% it was InsertTest that crashes test-container application. I think if you start running unit-tests in a loop for a night it will crash.

zabulus:
BTW, I think that test fails on AppVeyor CI is somehow connected to this issue.

zabulus:
> OT: Interesting the finalizer thread will not recover.

It shouldn't, it must crash an app. We have ```AppDomain.UnhandledException``` event handler. I think application doesn't crash because of it.

zabulus:
>	FirebirdSql.Data.FirebirdClient.dll!FirebirdSql.Data.Client.Native.FesDatabase.ParseStatusVector(System.IntPtr[] statusVector) Line 200	C#
 	FirebirdSql.Data.FirebirdClient.dll!FirebirdSql.Data.Client.Native.FesStatement.Free(int option) Line 626	C#
 	FirebirdSql.Data.FirebirdClient.dll!FirebirdSql.Data.Common.StatementBase.Release() Line 214	C#
 	FirebirdSql.Data.FirebirdClient.dll!FirebirdSql.Data.Client.Native.FesStatement.Release() Line 244	C#
 	FirebirdSql.Data.FirebirdClient.dll!FirebirdSql.Data.Client.Native.FesStatement.Dispose(bool disposing) Line 39	C#
 	FirebirdSql.Data.FirebirdClient.dll!FirebirdSql.Data.Common.StatementBase.Dispose() Line 131	C#
 	FirebirdSql.Data.FirebirdClient.dll!FirebirdSql.Data.FirebirdClient.FbCommand.Release() Line 840	C#
 	FirebirdSql.Data.FirebirdClient.dll!FirebirdSql.Data.FirebirdClient.FbCommand.Dispose(bool disposing) Line 101	C#
 	System.dll!System.ComponentModel.Component.~Component()	Unknown

cincuranet:
I'm fixing right now some GC related issues. And this looks very related. Not ready yet. But once that's done I believe this will be solved. 

Get back to it when I commit.

PR #24, opened 2015-02-05
Migrations

zabulus:
Please, take a look at my changes regarding to our project.


zabulus:
Here is still a couple of problems I haven't fixed yet:
1. Excess of maximum field, index, sequence, constraint name length. For now I've crutched it with hashing string. Looks like it works, but can't be sure for anything. Don't saying it looks ugly.
2. Refactored IIdentityBehavior. Added base class with all needed code. I'm sure we don't need to replicate create triggger/sequence SQL code in client's code. The behavior code needs only custom names.Still didn't found a suitable place to inject custom behavior to be dependent on database and not affect runtime.

PR #26, opened 2015-07-10
DNET-347 :  FbBatchExecution: Cannot determine SQL statement for ALTE…

Outlivier:
No unit tests provided, but if you execute the following code, we can check in table `RDB$CHARACTER_SETS` that default collations has changed :

```C#
FbScript fbScript = new FbScript("ALTER CHARACTER SET utf8 SET DEFAULT COLLATION unicode_ci_ai;\r\nALTER CHARACTER SET ISO8859_1 SET DEFAULT COLLATION FR_FR;");
fbScript.Parse();

FbBatchExecution fbBatch = new FbBatchExecution(new FbConnection(connString));
fbBatch.AppendSqlStatements(fbScript);
fbBatch.Execute(autoCommit: true);
```

### DIFF
--- a/NETProvider/src/FirebirdSql.Data.FirebirdClient/Isql/FbBatchExecution.cs
+++ b/NETProvider/src/FirebirdSql.Data.FirebirdClient/Isql/FbBatchExecution.cs
@@ -168,6 +168,7 @@ namespace FirebirdSql.Data.Isql
 				{
 					switch (statementType)
 					{
+						case SqlStatementType.AlterCharacterSet:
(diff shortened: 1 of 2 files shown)


cincuranet:
Great! Can you just place the `AlterCharacterSet` in `SqlStatementType.cs` before `AlterDatabase` to keep it organized. :)

zabulus:
>> we can check in table RDB$CHARACTER_SETS that default collations has changed
I think you can write a query to check this for possible unit-test.

PR #27, opened 2015-07-27
Add running Embedded server type tests

zabulus:


zabulus:
I've added running tests using embedded server configuration.

NETProvider/src/FirebirdSql.Data.UnitTests/TestsBase.cs
```diff
@@ -457,7 +453,17 @@ public static string BuildServicesConnectionString(bool includeDatabase)
 			return cs.ToString();
 		}
 
-		public static FbConnectionStringBuilder BuildConnectionStringBuilder()
+		public static IEnumerable<FbConnectionStringBuilder> BuildConnectionStringBuilder()
```
cincuranet: Why is here connection string for all `FbServerType`s? The `TestFixture` attributes should handle "all" the cases, isn't it?

NETProvider/src/FirebirdSql.Data.UnitTests/Properties/AssemblyInfo.cs
```diff
@@ -22,7 +22,7 @@
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
-[assembly: CLSCompliant(true)]
+[assembly: CLSCompliant(false)]
```
cincuranet: Can you elaborate?

NETProvider/src/FirebirdSql.Data.UnitTests/TestsBase.cs
zabulus: Yep, I've found TestFixture with arguments after I've made that changes. I'll rewrite this to use TestFixture as in TestBase descendants.

NETProvider/src/FirebirdSql.Data.UnitTests/Properties/AssemblyInfo.cs
zabulus: The compiler adds warning about ```[TestFixture(FbServerType.Embedded)```
See
https://msdn.microsoft.com/en-us/library/h9f682z5%28VS.80%29.aspx?f=255&MSPPError=-2147217396
This is because TestFixture attribute takes parameters as an array. (https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Attributes/TestFixtureAttribute.cs#L53)

zabulus: And I don't think that setting CLSCompliant to false for the unit-test changes anything.

NETProvider/src/FirebirdSql.Data.UnitTests/TestsBase.cs
```diff
@@ -109,9 +116,10 @@ public virtual void TearDown()
 				}
 			}
 			connection.Close();
+			connection.Dispose();
```
zabulus: Without this line, there is dangling database connection and database can't be dropped. 
I think it is a bug.
Reproduces only on GdsDatabase.

NETProvider/src/FirebirdSql.Data.UnitTests/Properties/AssemblyInfo.cs
cincuranet: Absolutely. I was just wondering.

NETProvider/src/FirebirdSql.Data.UnitTests/TestsBase.cs
cincuranet: Might be. Then the `Close` is not needed. `Dispose` will take care of that.

zabulus: Looks like it doesn't help. I have tests that failing if run they like "Run all" but they don't fail if run them one by one.

cincuranet: Looks like it's something else then.

zabulus: Looks like server keeps connection for some time after it's closed from client side.

zabulus:
I've narrowed down to FbConnectionTests.BeginTransactions* tests. Looks like there is something wrong with some of tests. They are failing randomly for now. If I add Ignore, other tests pass without problems.

zabulus:
Also there is couple Embedded specific fails:
ReadingCharOctetsTest - InvalidCastException. Unable to cast object of type 'System.String' to type 'System.Byte[]'.
ReadingVarcharOctetsTest - InvalidCastException. Unable to cast object of type 'System.String' to type 'System.Byte[]'.
SQLSTATETest - Expected: "42000" But was:  null

zabulus:
I've found the cause of random failing tests. I had 2.5.2 firebird default server.
I've updated the server and tests stopped failing.
I can revert workaround about unique database file names. But I think it will be good to keep the commit.

cincuranet:
I think the database name should be defined. There are not two tests running in parallel, so it's easier to handle it, especially in CI scenario.

zabulus:
@cincuranet will you merge the PR? I'll squash the commits.

cincuranet:
Yes. Is there a reason to not keep the database in default location (as it was before)?

zabulus:
My FB server created a database in a %WINDIR%\System32 folder.
Looks like some configuration issue of FB the server instance. It looked into system user (service account) folder.
If you say I can revert ```AppDomain.CurrentDomain.BaseDirectory```.

zabulus:
Just double checked. Installed Fb 2.5.4 32 bit on 64 bit machine. Database is created in %WINDIR%\SysWOW64. Without ```AppDomain.CurrentDomain.BaseDirectory```

cincuranet:
Sure. Because you're probably running the server as Windows Service. In that case the `cwd` is the directory you mentioned. But you can manually start it in different directory or set the path in config.

zabulus:
I see. Should I revert it? 
My only concern is about using AppDomain.BaseDirectory is fb_backup tests. They are starting to search for backup file from the %HOMEDRIVE%. 99% this is C:. but I can configurate the server to use another drive and test will eventually fail. Not to mention how long test is working while enumerates all system drive.
I prefer to leave database path near the unit-test dll.

cincuranet:
Yes. 

It will also fail if you use remote server. It's best of what we can do. These are in fact integration tests not unit tests. So some environment dependencies are expected.

zabulus:
I have doubts that anyone will be running tests on a remote server. Moreover ```localhost``` is predefined in the test.config as ```appsettings["Source"]```.
What if I add some environment variable to 'Database' and to backup file location. Let's say ```%TEMP%\TestFolder```. Server will create operational files in temp folder and searching of the backup file won't take ages. And test should work on remote sever.
What do you think.

cincuranet:
The backup tests will not work on remote server, because you can't check the backup is there because you don't have way to check the filesystem there. And I don't know any other better way to check the backup was created, at least some simple like the existence of the file.

This setup is plain simple. Database and backups are on the same location, which is the `cwd` for the server. I highly doubt anybody runs tests. Maybe just few people. And because I run tests using different versions and editions of FB, I'm starting it manually I always know where the files will be. If I need these files in a different directory I just start server elsewhere but use the same tests configuration.

zabulus:
Ok, looks like I got it.
What I suggest to do next:
1. I'll revert BaseDirectory from 'Database'
2. I rewrite test of backup/restore. It won't be two separate tests. I think existing of backup file we can check by trying to restore from it. 
What do you say?

zabulus:
Can you tell why CI don't run tests on build?
Maybe it will be good to make them running?

zabulus:
Ping

cincuranet:
The backup and restore tests are doing exactly this. But are separate tests. That's why they have weird names. So that's already done. 

I haven't had time to setup tests on CI yet. The matrix is quite huge. Feel free to contribute. I can commit the yml for AppVeyor later next week.

zabulus:
>> The backup and restore tests are doing exactly this. But are separate tests. That's why they have weird names. So that's already done.

Then I revert BaseDirectory.

>> The matrix is quite huge. 

Can you clarify this?

>>Feel free to contribute. I can commit the yml for AppVeyor later next week.

It will be great. Should I create another tracking issue for this?

cincuranet:
> Could you clarify this?

I'm running tests with different versions of Firebird (2.5, 2.1, ...) and editions (SS, SC, CS) and 32bit and 64bit.

> It will be great. Should I create another tracking issue for this?

We can keep it here. I don't have problem with that.

cincuranet:
The `yml` for AppVeyor is there. It's pretty basic. Just building. AppVeyor has no direct support for Firebird so it needs to be downloaded, unpacked and started before running tests.

zabulus:
Could you provide me permissions so I can cancel specified build on AppVeyor?

cincuranet:
You need to create your own account.

zabulus:
I'm stuck on the Default server tests
```TearDown : FirebirdSql.Data.FirebirdClient.FbException : unsupported on-disk structure for file C:\PROJECTS\FIREBIRDSQL-DATA-FIREBIRDCLIENT\NETPROVIDER_TESTS.FDB; found 0.2, support 11.2
  ----> FirebirdSql.Data.Common.IscException : unsupported on-disk structure for file C:\PROJECTS\FIREBIRDSQL-DATA-FIREBIRDCLIENT\NETPROVIDER_TESTS.FDB; found 0.2, support 11.2
   at FirebirdSql.Data.FirebirdClient.FbConnection.CreateData
```
@cincuranet do you have any ideas why does it fail?

cincuranet:
Not from top of my head.

zabulus:
I've reproduced test fails locally.
Tests don't fail if run from VS. But fail if run from nunit-console.exe.
Preliminary cause is that drop database doesn't delete test database file.
Will continue research.

cincuranet:
Interesting. I'm not running tests from VS at all, although I'm using the `nunit.exe` (the "GUI") and agent. I'll try `nunit-console.exe` later.

zabulus:
@cincuranet is it OK if I change tests to create/drop database in testfixture TearUp/TearDown. But instead I'll add cleaning all tables to test TearDown.
It wouldn't recreate database for every test (test time execution should decrease), but still would have enough test independency.

cincuranet:
Sure.

zabulus:
Looks like this change had fixed fails of Default tests.
Still there are such failing tests:
* FbCommandTests(Embedded).ReadingCharOctetsTest - it's a different representing of select cast('{0}' as char(10) need to be fixed separately
* FbCommandTests(Embedded).ReadingVarcharOctetsTest - same as above
*FbDatabaseSchemaTests(Default/Embedded).Tables - returns 7, but I create only 4. To be fixed in this PR
* FbExceptionTests(Embedded).SQLSTATETest - different behavior for embedded/default. In embedded actual is null. To be fixed separately.
* GuidTests(Default/Embedded) - to be fixed in this PR.
Also I've moved some tests to "Local" category and excluded from running.
* ConnectionPoolingTimeOutTest, DNET595. Produces wrong actual values. Will research it. Maybe I just patch to new values.
* BackupRestore_A_Backup01Test, BackupRestore_A_Backup02StreamingTest, BackupRestore_B_Restore02StreamingTest. Because they calling function GetBackupRestoreFullPath they all stuck the test run, and it breaks on 40 min timeout.
* ShutdownOnline2Test, ShutdownOnlineTest. Totally hungs. Haven't researched yet.
Everything is reproducable locally if run with nunit-console as specified in appveyor.yml.

zabulus:
@cincuranet please review this. And take a look at tests marked as Local category.

zabulus:
fixed embedded test fails.

zabulus:
Fixed backup tests.

cincuranet:
I see you made a lot of changes along the way. It's pretty hard to review all of these. Would it be possible to split this to multiple PRs, each one doing only one thing?

zabulus:
https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/pull/28
https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/pull/29
https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/pull/30
https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/pull/31

cincuranet:
Awesome!

PR #29, opened 2015-08-12
Tests running for Embedded

zabulus:


zabulus:
I've added fix for UnicodeTest fail. from build: https://ci.appveyor.com/project/cincura_net/firebirdsql-data-firebirdclient/build/0.0.0.260/job/esht3vq5cp107aqo/tests

Looks like when I've separated #27 I've reverted fix.

NETProvider/src/FirebirdSql.Data.UnitTests/FbCommandTests.cs
```diff
@@ -730,6 +756,15 @@ public void NoCommandPlanTest()
 				Assert.DoesNotThrow(() => { plan=cmd.CommandPlan; });
 				Assert.IsEmpty(plan);
 			}
+			}
+			finally
+			{
+				using (var cmd = Connection.CreateCommand())
+				{
+					cmd.CommandText = "drop table NoCommandPlanTest";
```
cincuranet: The `TearDown` would take care of that, no? IIRC you wanted to just cleanup the database to save time dropping and creating every time.

zabulus: ```TearDown``` cares only about tables that are created in ```SetUp```.


I can move ```NoCommandPlan``` table to ```SetUp/TearDown``` but the table is used locally in this test.

cincuranet: My idea was to drop all tables, procedures, triggers, ... and create these again. Fresh database. So whatever was done there is cleaned up. And you don't think about it. 

But maybe it's not that fast at the end. Your idea with cleaning up just the data might be faster.

zabulus: For this particular case. 
Do ```cmd.CommandPlan``` property need ```cmd.ExecuteNonQuery();``` to be called?
I've replaced it with Prepare and test passes and drop table not needed.

NETProvider/src/FirebirdSql.Data.UnitTests/FbConnectionStringTests.cs
```diff
@@ -27,8 +27,7 @@
 
 namespace FirebirdSql.Data.UnitTests
 {
-	[TestFixture]
-	public class FbConnectionStringTests : TestsBase
+	public class FbConnectionStringTests
```
cincuranet: Why is the `[TestFixture]` removed?

NETProvider/src/FirebirdSql.Data.UnitTests/FbDatabaseSchemaTests.cs
```diff
@@ -200,7 +202,7 @@ public void Tables()
 
 			tables = Connection.GetSchema("Tables", new string[] { null, null, null, "TABLE" });
 
-			Assert.AreEqual(3, tables.Rows.Count);
+			Assert.AreEqual(4, tables.Rows.Count);
```
cincuranet: Change from `3` to `4`?

NETProvider/src/FirebirdSql.Data.UnitTests/FbConnectionStringTests.cs
zabulus: reverted this.
only inheritance from TestsBase should be deleted

NETProvider/src/FirebirdSql.Data.UnitTests/FbServicesTests.cs
```diff
@@ -32,13 +32,13 @@
 
 namespace FirebirdSql.Data.UnitTests
 {
-	[TestFixture]
+	[TestFixture(FbServerType.Default)]
```
cincuranet: Not for `Embedded`?

zabulus: I had doubts about services. 
There are some tests that I think doesn't have sense for embedded connection. I.e. Shutdown, all user, properties tests.
I can extract default connection only tests to separate fixture and enable the rest with embedded connection.
What do you think?

cincuranet:
Will be merged manually. Changes are needed.

cincuranet:
Wondering whether we can force NUnit to group tests differently? Currently it's like this:
image
So running all `Default` or all `Embedded` tests is difficult. Categories might help, but that's just a lame solution.

zabulus:
in VS test explorer I'm using search string ```class:(Embedded)/class:(Default)```.

NETProvider/src/FirebirdSql.Data.UnitTests/TestsBase.cs
```diff
@@ -31,9 +31,45 @@
...
 namespace FirebirdSql.Data.UnitTests
 {
+	[SetUpFixture]
+	public class FirstClass
+	{
+		[SetUp]
+		public void SetupDatabase()
+		{
+			Prepare(FbServerType.Default);
+			Prepare(FbServerType.Embedded);
```
cincuranet: Umm. This fails if you're trying to run just one test in `Default`. I think I need to come up with some on-demand initialization.

cincuranet: That was easy. :) https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/commit/263fe8c50c4c7c00c973ba3f71f2776db91ddb72

PR #30, opened 2015-08-12
Fix FbExceptionTests(Embedded).SQLSTATETest

zabulus:


NETProvider/src/FirebirdSql.Data.FirebirdClient/Common/IscException.cs
```diff
@@ -187,7 +187,7 @@ private void BuildSqlState()
 		{
 			IscError error = this.Errors.Find(e => e.Type == IscCodes.isc_arg_sql_state);
 			// step #1, maybe we already have a SQLSTATE stuffed in the status vector
-			if (error != null)
+			if (error != null && !string.IsNullOrEmpty(error.StrParam))
```
cincuranet: What's the idea behind it?

zabulus: ```error.StrParam``` is null when test launched in embedded config

cincuranet: Interesting. I'll check with engine guys. Just to make sure we're not loosing the content somewhere.

zabulus:
Reimplemented fix. Found out missing parsing of sql state.

PR #32, opened 2015-08-25
Struct to class

cincuranet:


zabulus:
What is the reason for this change?

cincuranet:
Check the mailing list.

zabulus:
Still didn't get a point, though I don't see much trouble. 
My only concern about public visibility of these types. If any library user uses them in his code, it would be a breaking change for him. 
I suggest to go further and refactor API of these types to be less public. 
```FbDatabasesInfo``` should have only internal ctor and maybe immutable.
Fields of ```FbServerConfig``` should have only internal setters, or in ctor field initializers nad read-only fields.
As for ```FbTransactionOptions``` I think it shouldn't be touched.

**Comment by cincuranet** [comment]
> My only concern about public visibility of these types. If any library user uses them in his code, it would be a breaking change for him. 

Agreed. This will probably result into not-minor version bump.

> FbDatabasesInfo should have only internal ctor and maybe immutable.
> Fields of FbServerConfig should have only internal setters, or in ctor field initializers nad read-only fields.

Yes. I have this in my tree already.

**Comment by cincuranet** [comment]
Looking at it even the FbDatabasesInfo.Databases could be IReadOnlyList, at the end. What do you think?

**Comment by zabulus** [comment]
Yes, good point. It is informational objects, they shouldn't have any public mutable fields/methods.

**Comment by cincuranet** [comment]
OK. #33

File: NETProvider/src/FirebirdSql.Data.FirebirdClient/Services/FbServerConfig.cs [hunk 1]
```diff
@@ -1,26 +1,29 @@
 /*
- *	Firebird ADO.NET Data provider for .NET and Mono 
- * 
- *	   The contents of this file are subject to the Initial 
- *	   Developer's Public License Version 1.0 (the "License"); 
- *	   you may not use this file except in compliance with the 
- *	   License. You may obtain a copy of the License at 
+ *	Firebird ADO.NET Data provider for .NET and Mono
```
zabulus: @cincuranet do you plan somehow to fix all these copyright comments across repo?

cincuranet: Honestly I would like to remove these. But that's probably not possible. I'll probably clean the spaces solution wide with simple PS script.

PR #33, opened 2015-08-26
Using IReadOnlyList.

cincuranet:


zabulus:
```IReadOnlyList``` type introduced only in .net 4.5. Thus fails appveyor build Release_40

**Comment by cincuranet** [comment]
On .NET 4.0 I opted for `IList`. Not best, but what you can do.

**Comment by zabulus** [comment]
Maybe ```_databases.AsReadOnly()```. Not perfect, at least it will throw exception.

**Comment by cincuranet** [comment]
Heck, why not.

**Comment by zabulus** [comment]
```C#
#if (NET_40)
		public IList<string> Databases
#else
		public IReadOnlyList<string> Databases
#endif
```
I don't like this define. It changes library interface for client code with different compile options and can cause problems during migration client code from 4.0 to 4.5.

cincuranet:
Do you have better idea? 

This at least breaks during compile time, so the developer will fix it. And frankly it's not a difficult change. It's the same story as it was with .NET 3.5. Some pieces were not there too.

zabulus:
The best idea is to have only 4.0 build. At least library will have bare minimum, that will suffice both 4.5 and 4.0 users.
Though, if you OK with this define, I won't mind too.

cincuranet:
The .NET 4.5 build is needed for EF, the schema version 3 was introduced there.

zabulus:
But EF is for project EntityFramework.Firebird. FirebirdClient is lower layer for EF6 and provider only for 5.

cincuranet:
We still have a support for old EF that was part of .NET Framework. Nobody probably uses it, because it's getting stale in favor of `EntityFramework` NuGet package. Maybe it doesn't even work, I don't know. :) But it's just kept there.

Given the .NET 4.5 and .NET 4.6 is an in-place update of .NET 4.0 running on all supported platforms it would be doable to drop .NET 4.0 support. But I pretty sure a lot of people would complain with excuses here and there about .NET 4.0. :D

zabulus:
Ah I understood. It is all about System.Data.Entity assembly in .net 4.0 that declared deprecated in favor EntityFramework assembly.
In .net 4.5 all classes from System.Data.Entity migrated to EntityFramework, so FirebirdClient can target only .net 4.0 without problems. IIUC.
I think there should be two different EF providers in different projects, Something like following dependency:
```
EF5 (4.0)    EF6 (4.5)
    \               /
     \             /
      CommonEF (4.0)
            |
    FirebirdClient (4.0)
```

cincuranet:
Not possible. First the pre-EF6 versions don't have any provider model. The ADO.NET provider is expected to have everything needed. And also EF5 can run on .NET 4.5 (with new schema version), but also .NET 4.0 (basically EFv4 + some goodies). Crazy days.

zabulus:
Sigh, messy.

cincuranet:
Yes. Although I hated how EF6 changed the provider model, at the end it's now much cleaner.

PR #34, opened 2015-08-26
DNET-532

zabulus:


NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbParameterCollection.cs
```diff
@@ -217,12 +217,12 @@ internal int IndexOf(string parameterName, int luckyIndex)
...
 			if (luckyIndex != -1 && luckyIndex < _parameters.Count)
 			{
-				if (_parameters[luckyIndex].InternalParameterName.Equals(normalizedParameterName, StringComparison.CurrentCultureIgnoreCase))
+				if (_parameters[luckyIndex].InternalParameterName.Equals(normalizedParameterName, StringComparison.OrdinalIgnoreCase))
 				{
 					return luckyIndex;
 				}
 			}
-			return _parameters.FindIndex(x => x.InternalParameterName.Equals(normalizedParameterName, StringComparison.CurrentCultureIgnoreCase));
+			return _parameters.FindIndex(x => x.InternalParameterName.Equals(normalizedParameterName, StringComparison.OrdinalIgnoreCase));
```
cincuranet: Why do you think the ordinal compare is correct?

NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbParameter.cs
```diff
@@ -459,9 +470,12 @@ private void SetFbDbType(object value)
...
 		{
-			return !string.IsNullOrEmpty(parameterName) && !parameterName.StartsWith("@")
-				? string.Format("@{0}", parameterName)
-				: parameterName;
+			if (string.IsNullOrEmpty(parameterName) || parameterName[0] == '@')
+			{
+				return parameterName;
+			}
+
+			return parameterName.PadLeft(parameterName.Length + 1, '@');
```
cincuranet: This is obscure code.

NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbParameter.cs
```diff
@@ -46,6 +46,7 @@ public sealed class FbParameter : DbParameter, ICloneable
 		private object _value;
 		private string _parameterName;
 		private string _sourceColumn;
+		private string _normalizedParameterName;
```
cincuranet: This is actually good idea. :+1:

NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbParameterCollection.cs
cincuranet: Ahh, you took lucky shot. Check http://firebird.1100200.n4.nabble.com/DNET-532-tp4640493p4640508.html .

cincuranet:
I polished it a little (86a4997018c73b134edb31fe5e4ba16c16a170ff) and it's merged.

NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbParameter.cs
zabulus: Indeed, but it is fastest version of string Prepend. string.Concat works slower.
Though NormalizeParameterName now called less frequently it can be replaced with anything.

zabulus:
I don't see changes near string.Equals in master. These changes reduce execution time of the case to 9-10 secs.

cincuranet:
`string.Equals` where?

zabulus:
FbParameterCollection.IndexOf

cincuranet:
Yes. Because the ordinal compare is not correct. It needs to respect culture.

zabulus:
Interesting. But existing SqlClient (and maybe others) doesn't respect culture in DbParameterCollection. I am from phone now, later I'll give a link to sources.

cincuranet:
In this case it doesn't matter what others do. We already had issue with this, so I know we need to respect the culture.

zabulus:
Do you have a test about culture dependent parameters?

cincuranet:
Not one I'm aware of.

zabulus:
I still don't want to pay for things that I don't use. I'm planning to do next things:
1. Add tests with culture specific parameter name. For example, Turkish-i problem.
2. Add check if parameter and collection have fully ASCII name. If parameters in collection and parameter in IndexOf are fully ASCII (I think it would be 90% cases), use StringComparison.OrdinalIgnoreCase. If IndexOf Parameter or any parameter in collection are not fully ASCII IndexOf fallbacks to using StringComparison.CurrentCultureIgnoreCase.
3. Also I want to add time-based test described in jira issue, to check execution time of inserts.
@cincuranet any suggestions?

cincuranet:
> 1. & 2.

Sure. Go ahead.

> 3. Also I want to add time-based test described in jira issue, to check execution time of inserts.

It's going to be difficult to account for differences in execution speeds depending on the machine.

zabulus:
It can be test relative to "insert by index" case execution time. Let's say "insert by name" should be not more than "by index" time * n.

cincuranet:
It will work in some margin of error, but even then the differences in executing different code - especially comparing like AppVeyor, which is very slow, and regular machine might be significant. Though don't take me wrong. Sure knock yourself out.

PR #35, opened 2015-08-30
DNET-532

zabulus:
1. I've added test for culture aware comparison of parameter names.
2. I've added fallback to ordinal comparison in case parameter names don't have unicode characters.
3. Also added a little improvement to Remove method. It removes double comparison via Contains before deleting element from list.

About second I have doubts, the implementation is not very clean. Though, for original case difference between "by index" vs "by name" is about 4 seconds.

### DIFF
--- a/NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbParameterCollection.cs
+++ b/NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbParameterCollection.cs
@@ -39,6 +39,7 @@ namespace FirebirdSql.Data.FirebirdClient
 		#region Fields
 
 		private List<FbParameter> _parameters;
+		private bool _collectionHasUnicodeParameterNames;
 
 		#endregion
 
@@ -90,14 +91,22 @@ namespace FirebirdSql.Data.FirebirdClient
 		{
 			get { return ((ICollection)_parameters).SyncRoot; }
 		}
-		
+
 		internal bool CollectionHasParameterWithUnicodeName
 		{
 			get
 			{
-				return _parameters.Any(x => x.IsUnicodeParameterName);
-            }
+				if (!ParameterNameFlagEvaluated)
+				{
+					_collectionHasUnicodeParameterNames = _parameters.Any(x => x.IsUnicodeParameterName);
+					ParameterNameFlagEvaluated = true;
+				}
+
+				return _collectionHasUnicodeParameterNames;
+			}
 		}
+
+		internal bool ParameterNameFlagEvaluated { get; set; }
 
 		#endregion
 
@@ -155,9 +164,10 @@ namespace FirebirdSql.Data.FirebirdClient
 			EnsureFbParameterAddOrInsert(value);
 
 			value.Parent = this;
-				_parameters.Add(value);
-				return value;
-			}
+			_parameters.Add(value);
+			ParameterNameFlagEvaluated = false;
+			return value;
+		}
 
 		public override int Add(object value)
 		{
@@ -239,7 +249,7 @@ namespace FirebirdSql.Data.FirebirdClient
 
 		public void Remove(FbParameter value)
 		{
-			if (!_parameters.Remove(value))
+			if(!_parameters.Remove(value))
 			{
 				throw new ArgumentException("The parameter does not exist in the collection.");
 			}
--- a/NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbParameterCollection.cs
+++ b/NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbParameterCollection.cs
@@ -29,6 +29,7 @@ using System.Globalization;
 using System.Linq;
 
 using FirebirdSql.Data.Common;
+using System.Text;
 
 namespace FirebirdSql.Data.FirebirdClient
 {
@@ -89,6 +90,14 @@ namespace FirebirdSql.Data.FirebirdClient
 		{
 			get { return ((ICollection)_parameters).SyncRoot; }
 		}
+		
+		internal bool CollectionHasParameterWithUnicodeName
+		{
+			get
+			{
+				return _parameters.Any(x => x.IsUnicodeParameterName);
+            }
+		}
 
 		#endregion
 
@@ -146,9 +155,9 @@ namespace FirebirdSql.Data.FirebirdClient
 			EnsureFbParameterAddOrInsert(value);
 
 			value.Parent = this;
-			_parameters.Add(value);
-			return value;
-		}
+				_parameters.Add(value);
+				return value;
+			}
 
 		public override int Add(object value)
 		{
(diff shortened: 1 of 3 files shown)


NETProvider/src/FirebirdSql.Data.UnitTests/FbParameterCollectionTests.cs
```diff
@@ -23,19 +23,9 @@
 
 namespace FirebirdSql.Data.UnitTests
 {
-	[TestFixture(FbServerType.Default)]
-	[TestFixture(FbServerType.Embedded)]
-	public class FbParameterCollectionTests : TestsBase
+	[TestFixture]
```
zabulus: For this tests connection is not necessary

NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbParameterCollection.cs
```diff
@@ -90,6 +91,21 @@ public override object SyncRoot
 			get { return ((ICollection)_parameters).SyncRoot; }
 		}
 
+		internal bool? _collectionHasParameterWithUnicodeName;
```
zabulus: these fields are internal for tests

NETProvider/src/FirebirdSql.Data.UnitTests/FbParameterCollectionTests.cs
```diff
@@ -49,6 +39,64 @@ public void AddTest()
...
+				Assert.AreNotEqual(-1, command.Parameters.IndexOf("turkishIParameter"));
+			}
+			finally
+			{
+				System.Threading.Thread.CurrentThread.CurrentCulture = curCulture;
+            }
+		}
+
+		[Test]
+		[Category("Run")]
```
cincuranet: `Run` category?

NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbParameterCollection.cs
cincuranet: Isn't it enough to have property available?

NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbParameterCollection.cs
```diff
@@ -239,13 +266,13 @@ public override void Insert(int index, object value)
 
 		public void Remove(FbParameter value)
 		{
-			if (!Contains(value))
+			if(!_parameters.Remove(value))
```
cincuranet: :+1:

NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbParameterCollection.cs
```diff
@@ -214,20 +231,30 @@ public override int IndexOf(string parameterName)
 
 		internal int IndexOf(string parameterName, int luckyIndex)
 		{
+			bool parameterUnicodeName = Encoding.UTF8.GetByteCount(parameterName) != parameterName.Length;
+			bool collectionHasParameterWithUnicodeName = CollectionHasParameterWithUnicodeName;
+			StringComparison usedComparison = StringComparison.OrdinalIgnoreCase;
+			if(parameterUnicodeName || collectionHasParameterWithUnicodeName)
```
cincuranet: Minor formatting mistake. :)

NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbParameterCollection.cs
zabulus: in tests I've also checked that flag is nullified correctly when parameters changed, the property doesn't allow such check. They both, field and property, are implementation details.

NETProvider/src/FirebirdSql.Data.UnitTests/FbParameterCollectionTests.cs
zabulus: Need to remove

NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbParameterCollection.cs
zabulus: could you clarify?

cincuranet: Space after `if`.

NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbParameterCollection.cs
cincuranet: I feel that checking that field is too much. For instance if we recompute the value with every change it will still be correct but the test will assume it's not.

zabulus: I don't want to recompute field on each change. I intentionally made computing the field only on-demand if ```IndexOf``` is called.

zabulus: I think I can make the property private, but the field still need to be internal for tests

cincuranet: > I don't want to recompute field on each change. I intentionally made computing the field only on-demand if IndexOf is called.

That was just an example. You may as well use two properties, one with value and one saying whether the value needs recomputing or not. What I'm saying is that the internal representation should not matter. Only the desired outcome of behavior.

I don't mind if you don't like it. I'll merge it anyway and then I'll change it. So it may as well stay as is.

> I think I can make the property private, but the field still need to be internal for tests

That's even worse.

zabulus: > That was just an example. You may as well use two properties, one with value and one saying whether the value needs recomputing or not. What I'm saying is that the internal representation should not matter. Only the desired outcome of behavior.

Yes that may work.

cincuranet:
Found a case where it fails:
```
var collection = new FbParameterCollection();
var parameter = new FbParameter() { ParameterName = "test" };
collection.Add(parameter);
Assert.IsFalse(collection.CollectionHasParameterWithUnicodeName);
parameter.ParameterName = "řčšřčšřčš";
Assert.IsTrue(parameter.IsUnicodeParameterName);
Assert.IsTrue(collection.CollectionHasParameterWithUnicodeName);
```

zabulus:
Your case lacks ```IndexOf``` call. Only then field and property are reevaluated.
I have doubts about all this optmization. The case you wrote is a good evidence that all this stuff with field/property is pretty obscure.

cincuranet:
> Your case lacks IndexOf call. Only then field and property are reevaluated.

It fails anyway. ;)
```
var collection = new FbParameterCollection();
var parameter = new FbParameter() { ParameterName = "test" };
collection.Add(parameter);
var dummy1 = collection.IndexOf("dummy");
Assert.IsFalse(collection.CollectionHasParameterWithUnicodeName);
parameter.ParameterName = "řčšřčšřčš";
var dummy2 = collection.IndexOf("dummy");
Assert.IsTrue(parameter.IsUnicodeParameterName);
Assert.IsTrue(collection.CollectionHasParameterWithUnicodeName);
```

> I have doubts about all this optmization. The case you wrote is a good evidence that all this stuff with field/property is pretty obscure.

I agree. But you started the PR, so I though I'll give it a try.

zabulus:
> It fails anyway. ;)
I got why it fails. I didn't consider the case, where parameter name can be changed outside collection. I think I'll revert adding the field instead use only property, so it at least will be correct.

zabulus:
Removed field. Property just evaluates ```.Any()```.
I think I can implement optimization for this in other way. Something like ```version``` in ```List``` implementation. It should incremented when underlying list is changed or any element's ```ParameterName``` property chaged.

cincuranet:
> Removed field. Property just evaluates .Any().

It would be interesting to see (aka measure) what we're gaining now. And how behaves with 1, 10, 100, ... parameters in collection. Maybe the `Any` + ordinal compare would not bring desired benefit over culture aware compare.

> I think I can implement optimization for this in other way. Something like version in List implementation. It should incremented when underlying list is changed or any element's ParameterName property chaged.

That's again on the edge.

zabulus:
For case described in the issue difference is about 7-8 secs. Looks better than culture aware comparison.

cincuranet:
I see a different values. Most visible it's in worst case scenario. When the collection contains only ASCII parameter names, then the `Any` takes a lot of time offsetting the ordinal compare. I see about two orders of magnitude difference on a 100 items collection with 100 compares.

zabulus:
Ok I've added new implementation version. 10.59 (IBN) vs 6.11 (IBI)
This introduces ```version``` field that increments when changes of internal list or if field ```ParameterName``` of any FbParameter instance change occurs.
Also I've added test around Parent field. There were several issues about it.
It was never set by a ```Collection``` and it wasn't checked in insert method.

BTW, I think that it's getter can freely be public, though it's not part of ```DbParameter``` contract.

cincuranet:
OK, let me first pick the `Parent` fixes. These are going in, for sure. I'll commit these in a minute.

Why the `version`? Isn't simple `bool` flag enough? You just need to know whether the value computed is stale or not.

zabulus:
@cincuranet something left for this?

cincuranet:
Yes. This:
> Why the version? Isn't simple bool flag enough? You just need to know whether the value computed is stale or not.

zabulus:
I've fixed it in last commit

cincuranet:
Ahh, sorry. I missed that.

PR #36, opened 2015-09-03
Improve marshalling performance

zabulus:
Improve marshalling by direct reading memory to fields.
I know it looks very ugly.
Our queries begun run x1.1-1.25 faster after this fix.

### DIFF
--- a/NETProvider/src/FirebirdSql.Data.FirebirdClient/Client/Common/XsqldaMarshaler.cs
+++ b/NETProvider/src/FirebirdSql.Data.FirebirdClient/Client/Common/XsqldaMarshaler.cs
@@ -1,4 +1,4 @@
-ï»¿/*
+/*
  *	Firebird ADO.NET Data provider for .NET and Mono
  *
  *	   The contents of this file are subject to the Initial
@@ -21,47 +21,25 @@
 
 using System;
 using System.Runtime.InteropServices;
-using System.Text;
-
 using FirebirdSql.Data.Common;
+using System.IO;
 
 namespace FirebirdSql.Data.Client.Common
 {
-	internal sealed class XsqldaMarshaler
+	internal static class XsqldaMarshaler
 	{
-		#region Static Fields
+		static int sizeofXSQLDA = Marshal.SizeOf(typeof(XSQLDA));
+		static int sizeofXSQLVAR = Marshal.SizeOf(typeof(XSQLVAR));
 
-		private static readonly XsqldaMarshaler instance = new XsqldaMarshaler();
+		#region · Methods ·
 
-		#endregion
-
-		#region Static Properties
-
-		public static XsqldaMarshaler Instance
-		{
-			get { return XsqldaMarshaler.instance; }
-		}
-
-		#endregion
-
-		#region Constructors
-
-		private XsqldaMarshaler()
-		{
-		}
-
-		#endregion
-
-		#region Methods
-
-		public void CleanUpNativeData(ref IntPtr pNativeData)
+		public static void CleanUpNativeData(ref IntPtr pNativeData)
 		{
 			if (pNativeData != IntPtr.Zero)
 			{
 				// Obtain XSQLDA information
-				XSQLDA xsqlda = new XSQLDA();
 
-				xsqlda = (XSQLDA)Marshal.PtrToStructure(pNativeData, typeof(XSQLDA));
+				XSQLDA xsqlda = (XSQLDA)Marshal.PtrToStructure(pNativeData, typeof(XSQLDA));
 
 				// Destroy XSQLDA structure
 				Marshal.DestroyStructure(pNativeData, typeof(XSQLDA));
@@ -72,21 +50,22 @@ namespace FirebirdSql.Data.Client.Common
 					IntPtr ptr1 = GetIntPtr(pNativeData, ComputeLength(i));
 
 					// Free	sqldata	and	sqlind pointers	if needed
-					XSQLVAR sqlvar = (XSQLVAR)Marshal.PtrToStructure(ptr1, typeof(XSQLVAR));
+					XSQLVAREmpty sqlvar = (XSQLVAREmpty)Marshal.PtrToStructure(ptr1, typeof(XSQLVAREmpty));
 
 					if (sqlvar.sqldata != IntPtr.Zero)
 					{
 						Marshal.FreeHGlobal(sqlvar.sqldata);
 						sqlvar.sqldata = IntPtr.Zero;
 					}
+
 					if (sqlvar.sqlind != IntPtr.Zero)
 					{
 						Marshal.FreeHGlobal(sqlvar.sqlind);
 						sqlvar.sqlind = IntPtr.Zero;
 					}
 
-					IntPtr ptr2 = GetIntPtr(pNativeData, ComputeLength(i));
-					Marshal.DestroyStructure(ptr2, typeof(XSQLVAR));
+					Marshal.DestroyStructure(ptr1, typeof(XSQLVAREmpty));
+					Marshal.DestroyStructure(ptr1, typeof(XSQLVAR));
 				}
 
 				// Free	pointer	memory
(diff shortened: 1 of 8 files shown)


NETProvider/src/FirebirdSql.Data.FirebirdClient/Client/Common/XsqldaMarshaler.cs
```diff
@@ -21,58 +21,33 @@
...
 using System.Runtime.InteropServices;
-using System.Text;
-
 using FirebirdSql.Data.Common;
+using System.IO;
 
 namespace FirebirdSql.Data.Client.Common
 {
-	internal sealed class XsqldaMarshaler
+	internal static class XsqldaMarshaler
```
zabulus: the static class doesn't affect performance, though good improvement.

NETProvider/src/FirebirdSql.Data.FirebirdClient/Client/Common/XsqldaMarshaler.cs
```diff
@@ -153,75 +132,104 @@ public IntPtr MarshalManagedToNative(Charset charset, Descriptor descriptor)
...
 		}
 
-		public Descriptor MarshalNativeToManaged(Charset charset, IntPtr pNativeData)
+		public static Descriptor MarshalNativeToManaged(Charset charset, IntPtr pNativeData)
 		{
 			return MarshalNativeToManaged(charset, pNativeData, false);
 		}
 
-		public Descriptor MarshalNativeToManaged(Charset charset, IntPtr pNativeData, bool fetching)
+		static int sqlTypeOffset = Marshal.OffsetOf(typeof(XSQLVAR), "sqltype").AsInt();
```
zabulus: Now this code is very error-prone.

NETProvider/src/FirebirdSql.Data.FirebirdClient/Client/Common/XsqldaMarshaler.cs
```diff
@@ -153,75 +132,104 @@ public IntPtr MarshalManagedToNative(Charset charset, Descriptor descriptor)
...
+		static int ownernameOffset = Marshal.OffsetOf(typeof(XSQLVAR), "ownername").AsInt();
+		static int aliasname_lengthOffset = Marshal.OffsetOf(typeof(XSQLVAR), "aliasname_length").AsInt();
+		static int aliasnameOffset = Marshal.OffsetOf(typeof(XSQLVAR), "aliasname").AsInt();
+
+
+		static void MarshalVAR(IntPtr ptr, XSQLVAR var)
 		{
-			// Obtain XSQLDA information
-			XSQLDA xsqlda = new XSQLDA();
+			unsafe
```
zabulus: I thought maybe use NativeMemoryStream for this, WDT?

NETProvider/src/FirebirdSql.Data.FirebirdClient/Client/Common/XsqldaMarshaler.cs
```diff
@@ -229,25 +237,30 @@ public Descriptor MarshalNativeToManaged(Charset charset, IntPtr pNativeData, bo
...
 		}
 
-		private int ComputeLength(int n)
+		static int sizeofXSQLDA = Marshal.SizeOf(typeof(XSQLDA));
+		static int sizeofXSQLVAR = Marshal.SizeOf(typeof(XSQLVAR));
+
+		private static int ComputeLength(int n)
 		{
-			int length = (Marshal.SizeOf(typeof(XSQLDA)) + n * Marshal.SizeOf(typeof(XSQLVAR)));
+			var length = (sizeofXSQLDA + n * sizeofXSQLVAR);
```
zabulus: ```Marshal.SizeOf(typeof(XSQLDA))``` caching this value also good improve

NETProvider/src/FirebirdSql.Data.FirebirdClient/Client/Common/XsqldaMarshaler.cs
```diff
@@ -21,58 +21,33 @@
...
 				// Destroy XSQLVAR structures
-				for (int i = 0; i < xsqlda.sqln; i++)
+				for (var i = 0; i < xsqlda.sqln; i++)
 				{
-					IntPtr ptr1 = GetIntPtr(pNativeData, ComputeLength(i));
+					var ptr1 = GetIntPtr(pNativeData, ComputeLength(i));
 
 					// Free	sqldata	and	sqlind pointers	if needed
-					XSQLVAR sqlvar = (XSQLVAR)Marshal.PtrToStructure(ptr1, typeof(XSQLVAR));
+					var sqlvar = (XSQLVARCleanup)Marshal.PtrToStructure(ptr1, typeof(XSQLVARCleanup));
```
cincuranet: What about creating `XSQLVAREmpty` (or something like that) and from that deriving `XSQLVAR`? Then all the `sqltype` etc. fields will not be repeated and it can be just one place.

NETProvider/src/FirebirdSql.Data.FirebirdClient/Client/Common/XsqldaMarshaler.cs
cincuranet: Instead of `unsafe` block?

zabulus: I've mistaken. UnmanagedMemoryStream.
https://msdn.microsoft.com/en-us/library/system.io.unmanagedmemorystream(v=vs.110).aspx
Unsafe block won't go away, because stream needs byte*, fixing IntPtr won't go away. But it could fix maintainability and reduce risks in changing this code.
I'll try the fix and wrote about results.

NETProvider/src/FirebirdSql.Data.FirebirdClient/Client/Common/XsqldaMarshaler.cs
zabulus: yes, good point

NETProvider/src/FirebirdSql.Data.FirebirdClient/Client/Common/XsqldaMarshaler.cs
cincuranet: I still don't get what are you going to do with `UnmanagedMemoryStream`. :) Maybe because I think the below code is OK. It's readable. And it's just one place.

zabulus: UMS + BinaryReader and sequential rading fields.

cincuranet: Not sure it will be more readable etc. But let's see how it looks. There's nothing to loose. I might be biased because I'm looking at this code.

NETProvider/src/FirebirdSql.Data.FirebirdClient/Client/Common/XsqldaMarshaler.cs
```diff
@@ -169,59 +151,74 @@ public IntPtr MarshalManagedToNative(XSQLDA xsqlda, XSQLVAR[] xsqlvar)
...
 		}
 
-		public Descriptor MarshalNativeToManaged(Charset charset, IntPtr pNativeData, bool fetching)
+		static void MarshalXSQLVAR(IntPtr ptr, XSQLVAR var)
 		{
-			// Obtain XSQLDA information
-			XSQLDA xsqlda = new XSQLDA();
+			unsafe
+			{
+				using (BinaryReader reader = new BinaryReader(new UnmanagedMemoryStream((byte*)ptr.ToPointer(), sizeofXSQLVAR)))
```
cincuranet: I don't know man. I liked the previous version better.

NETProvider/src/FirebirdSql.Data.FirebirdClient/Client/Common/XsqldaMarshaler.cs
```diff
@@ -72,13 +50,14 @@ public void CleanUpNativeData(ref IntPtr pNativeData)
 					IntPtr ptr1 = GetIntPtr(pNativeData, ComputeLength(i));
 
 					// Free	sqldata	and	sqlind pointers	if needed
-					XSQLVAR sqlvar = (XSQLVAR)Marshal.PtrToStructure(ptr1, typeof(XSQLVAR));
+					XSQLVAREmpty sqlvar = (XSQLVAREmpty)Marshal.PtrToStructure(ptr1, typeof(XSQLVAREmpty));
```
cincuranet: What's the benefit of not using "full" `XSQLVAR`?

NETProvider/src/FirebirdSql.Data.FirebirdClient/Client/Common/XsqldaMarshaler.cs
cincuranet: On the other hand now the offsets are not needed.

NETProvider/src/FirebirdSql.Data.FirebirdClient/Client/Common/XsqldaMarshaler.cs
zabulus: XSQLVar class had array fields, and using XSQLVAREmpty in this place prevents allocation of that fields.

cincuranet: In this version there's no allocation in `XSQLVAR` (the first version had `new byte[32]` there, but no more - now it's handled by the `BinaryReader`).

zabulus: I took a closer look at all this. Results in latest commit.
I think it is better to have two classes as now. First for manual native-managed marshalling nad managed-native marshalling, and second only for cleanup HGlobal fields. If in this place use XSQLVar Marshal class allocates byte array fields and initializes them.

zabulus:
squashed and rebased

NETProvider/src/FirebirdSql.Data.FirebirdClient/Client/Common/XsqldaMarshaler.cs
cincuranet: Ahh, I see.

Just out of curiosity, what if there will be another `IntPtr` at the end in the future. Then the "empty" approach would not work, right?

cincuranet:
Merged by ab50f415286b23955995b26146ef9ee74cb6b9e1.

NETProvider/src/FirebirdSql.Data.FirebirdClient/Client/Common/XsqldaMarshaler.cs
zabulus: yes. It will need to add some 'trail' struct and calculate its offset related to XSQLVar begin, and only then call Marshal.PtrToStructure.

PR #37, opened 2015-09-11
Clean XSQLVAR cleaning

cincuranet:


cincuranet:
Should be good to go, right @zabulus?

zabulus:
Looks good.

PR #38, opened 2015-09-12
[WIP] Expose Async methods for FbCommand

zabulus:
During exploitation of Async EF queries in our project, I've found out that every query runs in sync when use Fb .NET provider. Indeed, https://github.com/Microsoft/referencesource/blob/master/System.Data/System/Data/Common/DBCommand.cs#L219 DbCommand returns ```Task.FromResult(non-async call)``` if async overload is not provided.

This is initial commit for adding Async methods to ```FbCommand``` class. It only uses already exposed Begin/End methods already exists in FbCommand via ```Task.Factory.FromAsync```.
I'm planning to do next things:
1. Add using CancellationToken in async methods, so they could be trully cancellable.
2. Replace ```Task.Factory.StartNew``` calls with API async calls if exists or starting brand new thread.
3. Cache delegate for BeginExecuterReader2.

### DIFF
--- a/NETProvider/src/FirebirdSql.Data.FirebirdClient/Common/TaskHelpers.cs
+++ b/NETProvider/src/FirebirdSql.Data.FirebirdClient/Common/TaskHelpers.cs
@@ -1,0 +1,18 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace FirebirdSql.Data.Common
+{
+	internal static class TaskHelpers
+	{
+		public static Task<TResult> FromException<TResult>(Exception exc)
+		{
+			var tcs = new TaskCompletionSource<TResult>();
+			tcs.SetException(exc);
+			return tcs.Task;
+		}
+	}
+}
--- a/NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbCommand.cs
+++ b/NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbCommand.cs
@@ -502,19 +502,31 @@ namespace FirebirdSql.Data.FirebirdClient
 
 			return RecordsAffected;
 		}
+
 #if NET_45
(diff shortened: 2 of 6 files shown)


zabulus:
@cincuranet PTAL.

cincuranet:
Calling `BeginInvoke` (or `FromAsync` on top of it) has nothing to do with real async. It just puts the execution onto some TP thread. No gain. The `BeginInvoke` here was a quick fix to be able to cancel command.

The async needs to go all the way up from socket, so no thread is blocked. I already started a little in branch `service-async`. But it will take some time. I'd like to get week or two of just this and nothing else.

zabulus:
That's what I supposed to do in 2. Replace Task.Factory.StartNew calls with API async calls if exists.

> Calling BeginInvoke (or FromAsync on top of it) has nothing to do with real async.

It is not async yes, just runs from another thread + FbCommand.Cancel command API but why don't use this? Look at my problem: I need a query to run in another thread and a possibility to cancel it via CTS. With this changes I can acomplish all of this. 
You already exposed Begin/End methods, why don't wrap them in Async methods, and in meantime continue making it non-blocking?

> The async needs to go all the way up from socket, so no thread is blocked. 

For Gds it can be achieved, but what about Fes?

cincuranet:
> It is not async yes, just runs from another thread + FbCommand.Cancel command API but why don't use this? Look at my problem: I need a query to run in another thread and a possibility to cancel it via CTS. With this changes I can acomplish all of this. 
You already exposed Begin/End methods, why don't wrap them in Async methods, and in meantime continue making it non-blocking?

I don't want to give false impression of supporting async. [That's worse than not supporting it](http://blog.cincura.net/233433-amazon-aws-net-sdk-knows-how-to-implement-asynchronous-operations-not/). Spinning the operation in another thread is super easy and you know _you're_ doing the dirty work. And yes, the hack with BeginXxx and EndXxx was a mistake. :\

I just wish to have a small uninterrupted time span for it.

> For Gds it can be achieved, but what about Fes?

Fes will be always synchronous. There's inherently nothing asynchronous.

PR #41, opened 2015-10-07
Add exception message on clearing running connections from pool

zabulus:


NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbConnectionPoolManager.cs
```diff
@@ -184,8 +184,10 @@ void CleanConnectionsImpl()
 					},
 					() =>
 					{
-						foreach (var item in _busy)
-							item.Dispose();
+						if(_busy.Any())
```
cincuranet: Wrong formatting.

NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbConnectionPoolManager.cs
```diff
@@ -184,8 +184,10 @@ void CleanConnectionsImpl()
...
+						{
+							throw new InvalidOperationException("Running connections exist.");
```
cincuranet: "Running"? What's that?

NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbConnectionPoolManager.cs
cincuranet: Shouldn't there be also an `else` clause?

cincuranet:
Don't forget to put it into tracker as I told you in list.

NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbConnectionPoolManager.cs
zabulus: What there should be?

NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbConnectionPoolManager.cs
zabulus: I'll fix that, just added the PR, to check test runs

zabulus:
> Don't forget to put it into tracker as I told you in list.
http://tracker.firebirdsql.org/browse/DNET-638

I'm going to add tests that expect the exception. And fix the exception message

cincuranet:
Reading the documentation for `SqlClient`:

> ClearAllPools resets (or empties) the connection pool. If there are connections in use at the time of the call, they are marked appropriately and will be discarded (instead of being returned to the pool) when Close is called on them.

Seems to be like an interesting idea.

PR #42, opened 2015-10-14
Pool without throwing exceptions

mrgleba:
My proposal follwoing our discussion
on the mailing list.

### DIFF
--- a/NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbConnectionPoolManager.cs
+++ b/NETProvider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbConnectionPoolManager.cs
@@ -105,7 +105,7 @@ namespace FirebirdSql.Data.FirebirdClient
 
 					var connection = _available.Any()
 						? _available.Pop().Connection
-						: CreateNewConnectionIfPossibleImpl(_connectionString);
+						: CreateNewConnection(_connectionString);
 					connection.SetOwningConnection(owner);
 					_busy.Add(connection);
 					return connection;
@@ -119,11 +119,15 @@ namespace FirebirdSql.Data.FirebirdClient
 					CheckDisposedImpl();
 
 					var removed = _busy.Remove(connection);
-					if (removed)
+					if (removed && _available.Count < _connectionString.MaxPoolSize)
 					{
 						_available.Push(new Item(DateTimeOffset.UtcNow, connection));
 					}
-				}
+                    else
+                    {
+                        connection.Dispose();
+                    }
+                }
 			}
 
 			public void CleanupPool()
@@ -193,13 +197,6 @@ namespace FirebirdSql.Data.FirebirdClient
 			{
 				if (_disposed)
 					throw new ObjectDisposedException(typeof(Pool).Name);
-			}
-
-			FbConnectionInternal CreateNewConnectionIfPossibleImpl(FbConnectionString connectionString)
-			{
-				if (_busy.Count() + 1 > connectionString.MaxPoolSize)
-					throw new InvalidOperationException("Connection pool is full.");
-				return CreateNewConnection(connectionString);
 			}
 		}
 


mrgleba:
I've jest run into a problem with the connection pool.
When I reach the limit an exception is thrown.

It seems to me as an suboptimal thing to di in such a case.

Con pool is there to gain speed and random exceptions are IMHO too high a price to pay.

I think in such a case new connections should be opened and added to the pool.
connectionString.MaxPoolSize should decide whether the connection is returned to the pool
or closed.
I would rather also test connectionString.MaxPoolSize with the "_available" stack
and not the "_busy" list. This would make connectionString.MaxPoolSize mean "how
many connections you would like to have at hand when you're working with the database"
rather than "how many concurrent connections would you like to allow", which
I think is closer to the goal of the pool.

Such behaviour would be much more predictable to the user.

cincuranet:
This will allow the # of connections to grow infinitely. That's pretty dangerous thing to do...

mrgleba:
The user should be able to make as many connections as they wish.
This is the case when the pool is disabled.
The pool IMHO should only provide speedup by keeping connections open (that's its main goal
as I understand it) and not limit the possible number of connections.

If there are many more connections than PoolMaxSize the are simply disposed rather than
returned to the pool. The number of connections may be as large as necessary but the
pool won't grow beyond MaxPoolSize.

cincuranet:
The pool is there to speedup connection establishing with reasonable resources. And the maximum is there to cap resources. 

You may want to set `MaxPoolSize` to some really big number and have basically same behavior (the cleanup will then shrink the pool when the need passes).

mrgleba:
Setting MaxPoolSize to a big number is also a suboptimal solution.
With this connections would not be closed until connectionlifetime 
epires which would result in many connections lingering open after a burst
in concurrent connections.

In my solutions the extra connections get closed immediately and only
at most maxpoolsize are persisted as a speedup.

My point is that the pool is here to speed up and not to impose concurrency limits.
The pool definetely should allow for bursts in connection numbers because they always happen in real life.

cincuranet:
> Setting MaxPoolSize to a big number is also a suboptimal solution.
With this connections would not be closed until connectionlifetime 
epires which would result in many connections lingering open after a burst
in concurrent connections.

Obviously. 

> My point is that the pool is here to speed up and not to impose concurrency limits.
The pool definetely should allow for bursts in connection numbers because they always happen in real life.

No, that's your definition of pool. The maximum is there exactly to provide (upper) limit.

mrgleba:
But this leaves us with 2 bad choices:
 - set maxpoolsize high and waste resources (connections stay open till connectionlifetime)
 - set maxpoolsize low and have random exceptions
In the current implementation there is no good value to set maxpoolsize to in a production environment. 

Maybe you would consider having 2 options: soft and hard limit:
 - connections are created and not returned to the pool if above soft limit
 - an exception is thrown over a hard limit

I would greatly appreciate your help here. It really is unusable in prduction in the current state.

cincuranet:
Honestly I think that connection pool bigger than say 20 is a code smell. Also lifetime of 5 seconds should be plenty of time. So unless you changed the maximum and you're using the default maximum of 100 and hitting it, I think there's something bad in your code (or database).

The soft and hard limit is interesting idea, but not aligned with ADO.NET and would cause IMO a lot of confusion for people not digging through code.

mrgleba:
Our app logic works more or less as follows:
 - user selects some element in the UI and a request for let's say 7 pieces of information are sent
 - server recieves the request, spawns 7 threads, each with its own conenction/transactions to fetch the data.
 - there can be many users and waiting for data doesn't block UI so the user can select another element while the current request id still being processed, so you can get to over 100 fairly easy under burst load while on average you would only need ~10 connections on standby.

Currently we use a connectionlifetime of 60 sec - most pauses in user's interaction in UI are longer than 5 s.

In the current situation we'll have to keep a fork (with my pull request) that would allow for unlimited bursts without leaving the connections lingering.

cincuranet:
I suppose the fetching takes couple of milliseconds. With that the connection is back in pool in no time. So even if multiple users are interacting with server there's still bunch of connections available as it's rotating only the few used to process the request.

zabulus:
> there can be many users and waiting for data doesn't block UI so the user can select another element while the current request id still being processed, so you can get to over 100 fairly easy under burst load while on average you would only need ~10 connections on standby.

In this place you should cancel existing requests, or block UI until they finished. Now you reached pool cap, but if you won't limit user pssibility to spawn new requests, you'll reach cap of resources. And this scenario leads to situation where all resources are belong to only one user's requests.

cincuranet:
Thinking about it more, I can add event to connection pool. Something like `OnOverMaxPoolSize` where developer can allow going over the maximum or not in which case the regular exception would be thrown. It's more or less like setting `MaxPoolSize` bigger, only dynamically and based on some code not hardcoded value in connection string. The connection would go back to pool when closed as usual.

Not sure it brings reasonable benefit...

mrgleba:
I gave it some more thought.

We could define 3 controling functions:
Func<FbConnectionString, int, bool>  CanGet
which given the connection string and current _busy.Count would decide if we create a new connection or throw
This would be called in CreateNewConnectionIfPossibleImpl

Func<FbConnectionString, int, bool>  CanRelease
which given the connection string and current _available.Count would decide if we relase to the pool
or close
This would be called in ReleaseConnection

Func<FbConnectionString, int, int, int>  DoClean
which given the connection string and current _available.Count and current _busy.Count would decide how many connections should be in the _available
as this function would get executed every 2 sec from the timer it could collect usage data and potentially make a predictive pool size.
This would be called in CleanupPool

With those three we could easily save the current beahviuor and we introduce no breaking change
while giving a lot of room to tune he pool to one's needs.

I think the obove funcs should static be members of FbConnectionPoolManager.
In this case (FbConnectionPoolManager made public) it'd also made sense
to move ClearPool/ClearAllPools.

Alternatively these new methods could be static on the FbConnection.

What are your thoughts on the above solution?

I caught some other minor improvements that could be done:

_busy could be a Dict<int, FbConnectionInternal> with
connection.Database.Handle being the key (handle should be unique among open connections)

CleanConnectionsImpl - I think we could change Parallel.Invoke to Task.Run - the connections could be closed in the background - no need waiting.

_available could be some mutable doubly-linked list with connections acquired/released from the right and closed due to expiration/pool shrinking on the left.

Would you be willing to consider such a change?
I so I would gladly prepare a pull request.

mrgleba:
It seems github ate the method signatures:

Func&lt;FbConnectionString, int, bool&gt; CanGet
Func&lt;FbConnectionString, int, bool&gt; CanRelease
Func&lt;FbConnectionString, int, int, int&gt; DoClean

cincuranet:
Very likely not. FirebirdClient is a library not a "lego". And the pool is optional, if somebody needs very different behavior they can implement own.

mrgleba:
That's reasonable. I think we'll go this way.
Thank you for your help & input on the topic.

PR #47, opened 2015-12-11
Updated to NUnit3.

cincuranet:


zabulus:
Maybe fb.net should migrate tests to xunit? Not much change. But if you migrate why don't migrate to .net community approved test framework?

cincuranet:
> .net community approved test framework

What's that???

PR #48, opened 2015-12-21
WIP for building. Add test run for FB3 version.

zabulus:


cincuranet:
I was thinking about tests on FB3 before and there are two items I'd like to solve somewhat nicely. 

First adding another attribute on tests seems to me bit lame - maybe it's the easiest, but worth thinking about anyway. The problem is, that with these attributes it's difficult to manage the tests - running some particular, or adding new for "FB3.1". Even with just `Default` and `Embedded` it was a lot of stuff. Haven't found better solution yet (though found different, but maybe even worse :)).

The other stuff is with FB3. Because there's more auth and encryption modes it would be nice to tests these too. Again exploding the combination of tests (i.e. starting FB with different configuration files) seems to be bit lame.

I restarted the build, let's see what happens.

zabulus:
In our project different configurations we support through xunit's ```IClassFixture<BaseConfig>```. Example you can see in issue https://github.com/xunit/xunit/issues/719 (our case is Example2 in the issue description). Though it's still not perfect solution, without improvement described in the issue, because of bloatcode.

cincuranet:
I'm solving this for now by explicitly starting the tests with different Firebird (i.e. https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/blob/matrix/appveyor.yml). Maybe for the future I'll somewhat integrate the configuration, now it's all "hardcoded" in the archive the AppVeyor downloads. At least the tests are same only the ambient stuff changes.

PR #55, opened 2016-05-21
Octects charset parameters reading fix

cincuranet:


zabulus:
Will it break existing code?
I ask because we had some issues with octets. Now can't remember what, though.
Could you push some alpha nupkg to nuger.org for testing with new bits?

cincuranet:
It will not (if the code was correct).

This will become 5.0.1 in a few days. But you can grab a build from
AppVeyor if you want it right now.

PR #56, opened 2016-05-25
added NUnit3TestAdapter

livioc:
i added the nuget package NUnit3TestAdapter to the test project to allow to execute tests with visual studio, all other nunit runners will still work as before this update

note: i excluded the new dlls from the rep with gitignore to use the default nuget restore behaviour

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ obj
 DDEX/installer/out/
 DDEX/installer/in/
 Provider/deploy/out
+Provider/lib/NUnit3TestAdapter.3.0.10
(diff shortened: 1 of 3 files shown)


cincuranet:
What's the benefit of having it locally? Either you have it in VS already or you can also run it from console (which I do).

livioc:
that to use it from vs you need at least one test project that references it so i must readd it every time that i reset my local changes with git but i understand if you prefer to keep the project clean from a thing used only to simplify the work of the people that used it :)

it's nice to use it from vs to start a test with the debugger

cincuranet:
I'll leave it open for a while to see what others might think.

PR #57, opened 2016-05-25
a failing test for DNET-683 and other tests

livioc:
added some tests to check boolean support with v2.5 of FirebirdSQL, 
the failing test for DNET-683 is FbBooleanSupportV2_5Tests.ParametrizedSelectTest(false) with Embedded


cincuranet:
http://tracker.firebirdsql.org/browse/DNET-683

PR #58, opened 2016-06-23
Cache check for database charset is None

gerdus:
avoids dictionary lookup and equality check on each value read
10-15% improvement in my tests


gerdus:
also did same for IsOctetsCharset as it was also showing up in hot path

PR #59, opened 2016-11-15
GdsConnection: null-reference crash on Linux.

o7g8:
The reading USERNAME environment variable (which is not defined on Linux) is replaced with Environment.UserName.

Any call to FBConnection.Open() on Linux led to an ArgumentNullException thrown from  GdsConnection.UserIdentificationData() which was attempting to read USERNAME environment variable (which doesn't exist on UNIX systems).


o7g8:
OK, the solution doesn't work on .NET Core https://github.com/dotnet/corefx/issues/5407
Should we then look into USERNAME first, then try  USER and, in case none of them are defined, use "" or "none" as a failover?

cincuranet:
This has been already fixed [here](https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/commit/722e3d8db319ca16c1b49bd8737c737c544aa927). And http://tracker.firebirdsql.org/browse/DNET-716.

o7g8:
Cool, thank you!

PR #60, opened 2016-11-22
socket.SetKeepAlive throws PlatformNotSupportedException on non-windows platforms

smili:


zabulus:
It is better to fix the issue in SetKeepAlive extension method (here https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/blob/fbe63f58d1958c116bab66bb6cd2230976f009ac/Provider/src/FirebirdSql.Data.FirebirdClient/Common/Extensions.cs#L58). It not worth to pollute the socket configuration code with OS-specific error handling.

cincuranet:
Yes, I agree. That's why it's not yet merged. I have other pieces there
to finish and I'll then do it together.

cincuranet:
Done in a40b1a4b34f2c46743ac817c657cdb939c4c5e88.

PR #61, opened 2016-12-23
Added additional Constructor for FbBatchExecution

anuviswan:
Contructor for FBBatchExecution used to have a second parameter accepting the FBScript values. This was later changed in 5.6 version (i believe). Most of the code which was written prior to the change are rising compilation errors due to the same, with little info or reference AppendSQLStatements Method. Considering this is a small change, wouldn't it be better if the transition is carried out in steps, may be marking it obsolete first and then removing it completely.

### DIFF
--- a/Provider/src/FirebirdSql.Data.FirebirdClient/Isql/FbBatchExecution.cs
+++ b/Provider/src/FirebirdSql.Data.FirebirdClient/Isql/FbBatchExecution.cs
@@ -79,11 +79,24 @@ namespace FirebirdSql.Data.Isql
 			}
 		}
 
-		/// <summary>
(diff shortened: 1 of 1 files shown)


cincuranet:
Having extra constructor was not following _pit of success_ approach, so it was removed. For a while, in minor updates, it was still there with `Obsolete` attribute (https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/commit/480416b2606d3c60cf1de5b6b94ea8d8401a5c09#diff-b4ec111f3bc87a2b9c99339476de22c3R106).

anuviswan:
Thanks for the update

PR #63, opened 2017-01-06
Finalizers and locking

cincuranet:


cincuranet:
Review request: @zabulus, @gonline.

Provider/src/FirebirdSql.Data.FirebirdClient/Client/Managed/Version11/GdsStatement.cs
```diff
@@ -210,11 +194,8 @@ protected override void Free(int option)
 			if (FreeNotNeeded(option))
 				return;
 
-			lock (_database.SyncObject)
-			{
-				DoFreePacket(option);
-				(Database as GdsDatabase).DeferredPackets.Enqueue(ProcessFreeResponse);
-			}
+			DoFreePacket(option);
+			(Database as GdsDatabase).DeferredPackets.Enqueue(ProcessFreeResponse);
```
zabulus: Database is guaranteed to be GdsDatabase here. I suggest several things:
1. Change as to cast. If my first statement ever becomes false, exception thrown here would be more informative: wrong cast vs nre.
2. Add ```GdsDatabase Database { get; set; }``` private(?) property, and use it everywhere inside the class, instead public interface property.

Provider/src/FirebirdSql.Data.FirebirdClient/Client/Native/FbClientFactory.cs
```diff
@@ -74,39 +75,17 @@ public static IFbClient GetFbClient(string dllName)
 			}
 
 			IFbClient fbClient;
-
-			// First, try to get the IFbClient from the cache.
-			lock (cache)
+			if (cache.TryGetValue(dllName, out fbClient))
```
zabulus: It is better to use GetOrAdd method here:
https://msdn.microsoft.com/ru-ru/library/ee378676(v=vs.110).aspx

Provider/src/FirebirdSql.Data.FirebirdClient/Client/Native/FesBlob.cs
```diff
@@ -84,117 +84,101 @@ public FesBlob(IDatabase db, TransactionBase transaction, long blobId)
...
+					requested,
+					tmp);
 
-					if (segmentLength > 0)
-					{
-						segment.Write(tmp, 0, segmentLength > requested ? requested : segmentLength);
-					}
+				if (segmentLength > 0)
+				{
+					segment.Write(tmp, 0, segmentLength > requested ? requested : segmentLength);
```
zabulus: Isn't it an error if segmentLength > requested here?

Provider/src/FirebirdSql.Data.FirebirdClient/Client/Native/FesDatabase.cs
```diff
@@ -112,15 +112,7 @@ public IFbClient FbClient
 
 		public object SyncObject
```
zabulus: Is it really needed?

Provider/src/FirebirdSql.Data.FirebirdClient/Client/Native/FesDatabase.cs
```diff
@@ -152,27 +136,17 @@ public FesDatabase(string dllName, Charset charset)
...
-					_charset = null;
-					_serverVersion = null;
-					_statusVector = null;
-					_transactionCount = 0;
-					_dialect = 0;
-					_handle.Dispose();
-					_packetSize = 0;
-				}
-
+				Detach();
```
zabulus: Throws Exceptions

Provider/src/FirebirdSql.Data.FirebirdClient/Client/Native/FesTransaction.cs
```diff
@@ -94,35 +94,20 @@ public FesTransaction(IDatabase db)
...
-							_db = null;
-							_handle.Dispose();
-							_state = TransactionState.NoTransaction;
-							_statusVector = null;
-						}
-
-						_disposed = true;
-						base.Dispose(disposing);
-					}
+					Rollback();
```
zabulus: Throws exception?

Provider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbTransaction.cs
```diff
@@ -173,92 +145,77 @@ public override void Rollback()
...
+			catch (IscException ex)
+			{
+				throw new FbException(ex.Message, ex);
+			}
 		}
 
 		public void Commit(string savePointName)
 		{
 			EnsureSavePointName(savePointName);
-			lock (this)
```
zabulus: Just great you removed lock(this)

cincuranet: Yeah. It's removed everywhere. It was part of the motivation to fix the finalizers mess.

Provider/src/FirebirdSql.Data.FirebirdClient/Client/Managed/Version10/GdsEventManager.cs
```diff
@@ -60,18 +57,15 @@ public GdsEventManager(int handle, string ipAddress, int portNumber)
 
 		public void QueueEvents(RemoteEvent remoteEvent)
 		{
-			lock (this)
-			{
-				_events[remoteEvent.LocalId] = remoteEvent;
+			_events[remoteEvent.LocalId] = remoteEvent;
 
 #warning Jiri Cincura: I'm pretty sure this is a race condition.
```
cincuranet: I have already working piece where the thread is removed completely and non-blocking IO is used instead. Still some work left though. We'll test it heavily with @gonline in coming weeks.

Provider/src/FirebirdSql.Data.FirebirdClient/Client/Managed/SspiHelper.cs
```diff
@@ -326,27 +324,27 @@ public SspiHelper(string securityPackage, string remotePrincipal)
...
-			SecInteger expiry = new SecInteger(0);
+			_clientContext = new SecHandle();
+			SecInteger expiry = new SecInteger();
 			uint contextAttributes;
 			SecBufferDesc clientTokenBuf = new SecBufferDesc(MAX_TOKEN_SIZE);
 			try
 			{
 				int resCode = InitializeSecurityContext(
 					ref _clientCredentials,
 					IntPtr.Zero,
```
cincuranet: Yeah. Maybe in separate branch and completely across the whole code base.

Provider/src/FirebirdSql.Data.FirebirdClient/Client/Native/FesDatabase.cs
cincuranet: Not anymore. Removed.

Provider/src/FirebirdSql.Data.FirebirdClient/Client/Native/FesBlob.cs
cincuranet: Could be. I'm focusing here on finalizers. Maybe you can put it into tracker.

Provider/src/FirebirdSql.Data.FirebirdClient/Client/Managed/Version11/GdsStatement.cs
cincuranet: The `GdsDatabase` can be from `Version10` to `Version13` (currently), so it would have to be in every derived `GdsStatement`. I'd like to use generics somehow here. Not there yet though.

Provider/src/FirebirdSql.Data.FirebirdClient/Client/Native/FbClientFactory.cs
cincuranet: True. Fixed in e9f26ee.

Provider/src/FirebirdSql.Data.FirebirdClient/Client/Managed/Version10/GdsDatabase.cs
```diff
@@ -378,31 +326,28 @@ protected void ProcessCreateResponse(GenericResponse response)
...
+			{
+				try
 				{
-					try
-					{
-						CloseConnection();
-					}
-					catch
-					{ }
+					CloseConnection();
```
cincuranet: There's a bunch of these exception-swallowing pieces across the code base and I don't know whether I like it. I tend to think the exception should be propagated. It's only few pieces where swallowing makes sense. See also my 3. in PR reply.

cincuranet:
> But could you do them next time incrementally?

I wish. At one point Git got so confused with conflicts in rebase and following merge it was unable to continue. So I had to squash few times.

> I don't think that polluting Dispose with null assignment statements is good idea.
It's enough to dispose owning fields, that support dispose, set double-dispose prevention flag and call base.Dispose in finally block.

I agree. I left it there because it hurts nothing, as it was there before, right now and can (maybe) uncover some use-after-dispose thanks to NRE. Eventually I'll remove those.

Also the `base.Dispose` in finally... Not sure about that in general. Should the exception happen the finally block might talk to server with some corrupted state. I think this would need to be evaluated carefully later on.

>  

About the exceptions from `Dispose` calls. I think the exceptions should be propagated out. And once the exception happens you can't trust the state of the object anymore and shouldn't use it, IMO. So why pretend something else. Although exceptions from `Dispose` can be annoying. I have to do it and test it to see how the programming with it feels. If it feels utterly wrong, I'll think about something else.

zabulus:
> About the exceptions from Dispose calls. I think the exceptions should be propagated out. And once the exception happens you can't trust the state of the object anymore and shouldn't use it, IMO. So why pretend something else. Although exceptions from Dispose can be annoying. I have to do it and test it to see how the programming with it feels. If it feels utterly wrong, I'll think about something else.

Sure, they should be propagated. My comments regarding exceptions related to following code:
```
<Exception throwing function call>
_disposed = true;
```
if exception thrown before this code, double-dispose checking won't work, the flag won't be set.
I suggest just to put in in finally block.

> Also the base.Dispose in finally... Not sure about that in general. Should the exception happen the finally block might talk to server with some corrupted state. I think this would need to be evaluated carefully later on.

About this, if exception occurs in derived dispose, it's problem of cleaning derived code. Base code should also have opportunity to clean base-only stuff. 
I agree about additional evaluation, AFAIR, there could be issues with Release method. It's declared as virtual and called in Dispose, and it calls base.Release in the body.

> I wish. At one point Git got so confused with conflicts in rebase and following merge it was unable to continue. So I had to squash few times.

You could spawn PRs more frequently with smaller changes and use merges instead rebases. There won't be pretty history, but who cares :) if it just works.

Provider/src/FirebirdSql.Data.FirebirdClient/Client/Native/FesBlob.cs
zabulus: http://tracker.firebirdsql.org/browse/DNET-731

cincuranet:
> Sure, they should be propagated. My comments regarding exceptions related to following code:

> \<Exception throwing function call>
> _disposed = true;
> if exception thrown before this code, double-dispose checking won't work, the flag won't be set.
I suggest just to put in in finally block.

Ahh, then the `_dispose = true;` can be at the top. As long as the exception has been thrown the state of the object is unknown anyway and the object shouldn't be touched.

> About this, if exception occurs in derived dispose, it's problem of cleaning derived code. Base code should also have opportunity to clean base-only stuff.

In case of exception I don't thing the base class should have any opportunity. It might be best-effort approach, but that's all. 

> I agree about additional evaluation, AFAIR, there could be issues with Release method. It's declared as virtual and called in Dispose, and it calls base.Release in the body.

Yep, something like this. Or simply fucking up _object handle_ in derived class and then doing something on the server with wrong one in base class.

I don't think one-size-fits-all would work here. It needs to be evaluated one by one. Also. I'd like to remove all the `Close`, `Release` etc. methods and have single one `Dispose` and only that one is called at a known occurrences. Single entry point for "closing" stuff.

> You could spawn PRs more frequently with smaller changes and use merges instead rebases. There won't be pretty history, but who cares :) if it just works.

It needs to be one logical self-contained change. Especially if it's something like finalizers. Because 5 years later I might have to look into history for something and I (or somebody else) need to understand the context of the change.

cincuranet:
OK, I've put the `_disposed = true;` to the top. It's [here](https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/tree/disposed-first).

PR #64, opened 2017-02-09
Allocations in XdrStream 

jack128:
This pull request fix a few memory issues in XdrStream.Read() method. Quick test (selecting 200К records with one BIGINT field) shows 30% less GC collections and about ~5% better speed..


zabulus:
Isn't it manual implementation of what BufferedStream is doing? Have you considered using it?

jack128:
@zabulus 
Well, XdrStream is combination of functionality of BinaryReader, BufferedStream and sort of DecompressionStream.. I think using BinaryReader is OK. 
And BufferedStream requires seekable underlying stream, so we cannot use one directly.

zabulus:
>And BufferedStream requires seekable underlying stream, so we cannot use one directly.

I don't see such requirements. Main use-case of BufferedStream is wrapping NetworkStream, which AFAIK is not seekable.

> Well, XdrStream is combination of functionality of BinaryReader, BufferedStream and sort of DecompressionStream.. I think using BinaryReader is OK.

Yep, I understand, that XdrStream is overloaded with responsibilities. But instead to simplify it, your changes adds more complexity to it.

jack128:
> I don't see such requirements.
Oh, yes, it's my mistake. For non-compressed streams we can use BufferedStream + BinaryReader. 
But for compressed streams - I dont'n see good solution without manual buffering..

zabulus:
> But for compressed streams - I dont'n see good solution without manual buffering..

Agree, maybe it's worth to extract the compression logic to separate class, maybe derived. Just thoughts about this.

jack128:
Hm.  What's about  line endings ??  For example, WarningMessageCallback.cs use win style (0A0D), but ParameterBuffer.cs use unix style (0A)

cincuranet:
Who's talking about line endings?

jack128:
>> Who's talking about line endings?
OK, but what are you talking about ?? I dont see any docs about code style.
formatting defines in .editorconfig, and as I can see, I respect this settings

cincuranet:
General C# formatting. Look how the code is formatted in rest of the file.

cincuranet:
Replaced by #65.

PR #65, opened 2017-03-16
Improve input buffer

gerdus:
I looked at DNET-740 (http://tracker.firebirdsql.org/browse/DNET-740) and found the single threaded performance to be worse than it was. 
Main performance killer was `_inputBuffer.RemoveRange(0, data.Length);`

Replaced the list<byte> with a class that keeps a list of the readbuffers and removes them when no longer needed. 

Example performance on same data with 444528 rows
5.8: 11s
after this pull: 3.9s


zabulus:
Looks like a few tests are failed with your changes.

jack128:
This issue (and many others) fixed in https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/pull/64

gerdus:
Feel free to close this pull if @jack128 #64 is better and will be merged.
After my last commit those tests are passing locally but appveyer is failing with "Specify which project or solution file to use because the folder "C:\projects\firebirdsql-data-firebirdclient\Provider\src\FirebirdSql.Data.FirebirdClient" contains more than one project or solution file."

cincuranet:
Why is the `ReadBuffer` class having so much members and not much is used? Is it copied from somewhere?

Also it seems overly complex. When I was finishing compression support I remember thinking that the data could be just handled using one array and passed from there. Though I never finished it before the release.

gerdus:
All the methods in ReadBuffer is used except Take which I forgot to remove. Some methods are only called by one other method and so can be inlined if you prefer. The other thing that makes it appear more complex is that internally it operates one element at a time while the public api is buffer and length. I did try a version that internally also worked with a range but it was exponentially more complex.
I replaced the list of buffers approach with a ring buffer that internally uses one array.

The core logic comes from https://www.snellman.net/blog/archive/2016-12-13-ring-buffers/ (Array + index + length version).

cincuranet:
Maybe, blind guess - haven't tried replacing it myself - the `ArraySegment` might help here?

gerdus:
Simplified the ReadBuffer class. 

I also upgraded one of the projects(xproj to csproj) for use on VS2017 to fix the appveyor build failing.

cincuranet:
I looked at the `ReadBuffer` and looks fine to me. I'll not directly merge it, probably, I'll take it into the project and polish it little.

zabulus:
Have you considered using https://www.nuget.org/packages/System.Buffers/. There is ArrayPool, with exactly functions you need here: Rent/Return.

cincuranet:
@zabulus> I did. But it does not support .NET 4.0. Which I wouldn't mind dropping from FirebirdClient ;), but some people wouldn't like that move. Also taking dependency on a library makes it difficult for distribution (sadly some people are still not using NuGet).

zabulus:
> Also taking dependency on a library makes it difficult for distribution (sadly some people are still not using NuGet).

ILMerge can be used here. It's also published as nuget package and invoking could be added as publish step.

> I did. But it does not support .NET 4.0. Which I wouldn't mind dropping from FirebirdClient ;), but some people wouldn't like that move.

Yep, that's a problem. Maybe compiling it as .net 4.0 assembly worth a try.

cincuranet:
ILMerge sucks when you need debug with PDBs. Also the effort compiling for .NET 4 and merging can be easily offset by just using @gerdus's class, which works fine. Nothing more is needed.

cincuranet:
Merged by https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/commit/993517ee480c1c5e043c1183087982537c793078.

PR #66, opened 2017-03-22
- Fix memory leak in Firebird 3 on implicit transaction DNET-743

DiverDW:


cincuranet:
This doesn't seem right to me. Releasing the statement handle basically right after executing would prevent executing the same statement again, Can you explain on what is your fixed based?

cincuranet:
This a bug in Firebird server itself. More info in the [ticket](http://tracker.firebirdsql.org/browse/DNET-743).

PR #69, opened 2017-08-24
Upgrade to .net standard2.0

souchprod:
Hi, i made the project & little changes required to use the client under the latest fw.
The main benefit is the return of the DBNull support, really problematic under the 1.6 version.

AssemblyInfo.cs should not be merged and the signature is missing from the project.


Provider/src/FirebirdSql.Data.FirebirdClient/Properties/AssemblyInfo.cs
```diff
@@ -25,13 +25,13 @@
 [assembly: CLSCompliant(true)]
 [assembly: ComVisible(true)]
 [assembly: Guid("F37B8373-51FA-4CC6-934E-DF147DFBD9F1")]
-[assembly: AssemblyCompany("FirebirdSQL")]
-[assembly: AssemblyProduct("FirebirdClient")]
+[assembly: AssemblyCompany("SouchProd")]
+[assembly: AssemblyProduct("FirebirdClient for .NetStandart 2")]
```
Mahagon: There is a typo: "Standar**t**"

souchprod: True, i will fix'n commit

cincuranet:
Thanks for contribution. But I can't merge it. It's just a 1.6 rebuild as 2.0. It's not taking into account any improvements available in 2.0. Not mentioning changes that should not be there.

Also the NET Standard 2.0 support is already in [ns2 branch](https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/tree/ns2).

zabulus:
3rd party libraries should target a least .net standard. For such libraries as Fb.NET you need not increase target standard, but instead lowering it as much as it possible to be available on more platforms.

cincuranet:
Correct. I still do support 1.6. But also 2.0, because there's a bunch of new APIs provider can benefit from. And if somebody is on .NET Core 2 (or related platform) the .NET Standard 2.0 support is only a win.

zabulus:
Have you considered extracting all 2.0-dependent stuff to separate library so you won't be tied to support two versions of Fb.NET library, but to support two libraries?

cincuranet:
Why two versions? It's one version with multiple targets.

zabulus:
Sorry, by "two versions" I meant two versions of code, produced by ``ifdefs```. It's just suggestion to remove ```ifdef```'ed code and use it in separate extension assembly. On the other hand it will cause additional effort.

cincuranet:
That's not possible in all cases. Sometimes it's not code only, but also structure. For example from top of my head [this](https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/blob/ns2/Provider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbConnection.cs#L36).

souchprod:
Hi, i understand the rejection, sorry for that.

In my situation, this PR fix a blocking issue located in FbParameter.cs row 416:

`TypeCode code = Type.GetTypeCode(value.GetType());

(...)

    #if !NETSTANDARD1_6
	case TypeCode.DBNull:
    #endif`

This very small change by itself resolve a bunch of issue IMO, i was not able to find another way to fix it properley without targeting NS 2.0, do you see a better option ?

cincuranet:
That's exactly what [.NET Standard 2 support](https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/tree/ns2) solves.

souchprod:
Damned true, i completely missed this branch :"'(
Perfect then !

zabulus:
> That's not possible in all cases.

I have several ideas for this direction of changes. Do you interested in continuation of this discussion?

cincuranet:
@souchprod> Looking forward to it as well. I need to merge the branch and include one more change (http://tracker.firebirdsql.org/browse/DNET?report=com.atlassian.jira.plugin.system.project:roadmap-panel) and it's going to be released. Fingers crossed. :)

@zabulus> Sure. Feel free to continue here.

zabulus:
1. ```Serializable``` attribute is available from ```System.Runtime.Serialization.Formatters``` nuget package, which targets netstandart 1.4
2. ```Schema``` part could be all go to separate library.
3. As for structure changes, my general suggestion would be to have partial implementations for each class that needs additional structure. For your example it would be:
FbConnection.cs:
```
public sealed partial class FbConnection : DbConnection
```
FbConnection_NetStandart2.0.cs or FbConnection_ICloneable.cs (included only in projects that supports ICloneable):
```
public sealed partial class FbConnection : ICloneable
{ object ICloneable.Clone() }
```
EDIT: But on the second glance it's not extracting to separate assembly as I've suggested, it's just reorganizing existing ifdefs in another manner. Looks like without introducing breaking change it can't be accomplished.

cincuranet:
1. Interesting. Somehow I missed it. I added it now with [this commit](https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/commit/f88197612a45c28ed68de5c625c0e69f90e03b61).

2. True. But the missing `Schema` API was a great mistake by Microsoft (and they knew/know it). In .NET Standard 2.0 it's fixed. Could be also a breaking change for someone using `FirebirdSql.Data.FirebirdClient.dll` via some factory/injection.

3. Your edit sums it up. :) On the other hand I'm hoping in coming months to clean the projects and ifdefs a lot. Together with multitargeting and `NET40` drop.

PR #71, opened 2017-09-01
Solution for DNET-770 issue.

ralmsdeveloper:
INSERT (...) RETURNING (...) statement:
* Works fine when using ExecuteScalar ()
* Does not work when using Executereader ()

http://tracker.firebirdsql.org/browse/DNET-770


souchprod:
You duplicated the PR pull request with your latest commit isn't it ? 0_o

cincuranet:
Yep. The a5823e8e948c759196e32ba1d93f3a5ec6ae7f78 should be removed.

PR #72, opened 2017-09-13
Start of implementation Migration

ralmsdeveloper:
Implementation Migration


cincuranet:
Look at [this](https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/blob/master/Provider/src/EntityFramework.Firebird/DefaultFbMigrationSqlGeneratorBehavior.cs#L29) class. We should reuse it as people might have already written it.

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Migrations/FbMigrationsSqlGenerator.cs
```diff
@@ -15,16 +15,377 @@
...
+			if (_options.Settings.IsSupportIdentityIncrement)
+				return;
+
+			foreach (var column in operation.Columns.Where(p => !p.IsNullable))
+			{
+				var colAnnotation = (IAnnotation)column.FindAnnotation(FbAnnotationNames.ValueGenerationStrategy);
+				if (colAnnotation != null)
+				{
+					var valueGenerationStrategy = colAnnotation.Value as FbValueGenerationStrategy?;
+					if (valueGenerationStrategy == FbValueGenerationStrategy.IdentityColumn
```
cincuranet: This should be for trigger not identity.

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Internal/FbOptions.cs
```diff
@@ -15,16 +15,28 @@
...
 	{
+
+		private Lazy<FbSettings> _fbSettings;
+		public virtual FbSettings Settings => _fbSettings.Value;
+
 		public void Initialize(IDbContextOptions options)
 		{
 			var fbOptions = GetOptions(options);
+
+			_fbSettings = new Lazy<FbSettings>(() => fbOptions.Connection != null ?
```
cincuranet: Formatting.

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Internal/FbOptions.cs
```diff
@@ -33,6 +45,6 @@ public void Validate(IDbContextOptions options)
 		}
 
 		static FbOptionsExtension GetOptions(IDbContextOptions options)
-			=> options.FindExtension<FbOptionsExtension>() ?? new FbOptionsExtension();
+			=> options.Extensions.OfType<FbOptionsExtension>().FirstOrDefault();
```
cincuranet: Why?

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Extensions/FbStringExtension.cs
```diff
@@ -0,0 +1,30 @@
...
+ *    All Rights Reserved.
+ */
+
+//$Authors = Jiri Cincura (jiri@cincura.net), Rafael Almeida (ralms@ralms.net)
+
+namespace FirebirdSql.EntityFrameworkCore.Firebird.Extensions
+{
+    public static class FbStringExtension
+    {
+	    public static string MaxLength(this string src, int maxLength)
```
cincuranet: This can be written with `Substring` directly. No need for extra method.

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Migrations/FbMigrationsSqlGenerator.cs
```diff
@@ -15,16 +15,377 @@
...
+				.Append("ALTER TABLE ")
+				.Append(Dependencies.SqlGenerationHelper.DelimitIdentifier(operation.Table, operation.Schema))
+				.Append(" DROP CONSTRAINT ")
+				.Append(Dependencies.SqlGenerationHelper.DelimitIdentifier(operation.Name))
+				.AppendLine(Dependencies.SqlGenerationHelper.StatementTerminator);
+
+			EndStatement(builder);
+		}
+
+		protected override void Generate(
```
cincuranet: Formatting.

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Migrations/FbMigrationsSqlGenerator.cs
```diff
@@ -15,16 +15,377 @@
...
+			bool nullable,
+			object defaultValue,
+			string defaultValueSql,
+			string computedColumnSql,
+			IAnnotatable annotatable,
+			IModel model,
+			MigrationCommandListBuilder builder)
+		{
+
+			var Identity = false;
```
cincuranet: Variable name.

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Migrations/FbMigrationsSqlGenerator.cs
```diff
@@ -15,16 +15,377 @@
...
+				.Append(Dependencies.SqlGenerationHelper.DelimitIdentifier(operation.Table, operation.Schema))
+				.Append(" ALTER ")
+				.Append(Dependencies.SqlGenerationHelper.DelimitIdentifier(operation.Name))
+				.Append(" TO ")
+				.Append(Dependencies.SqlGenerationHelper.DelimitIdentifier(operation.NewName));
+
+			EndStatement(builder);
+		}
+
+		protected override void ColumnDefinition(
```
cincuranet: Formatting.

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Storage/Internal/FbSqlGenerationHelper.cs
```diff
@@ -15,14 +15,54 @@
...
+		public FbSqlGenerationHelper(RelationalSqlGenerationHelperDependencies dependencies, IFbOptions options)
 			: base(dependencies)
-		{ }
+		{
+			_options = options;
+		}
+
+		public override string EscapeIdentifier(string identifier)
+		{
+			return identifier.MaxLength(_options.Settings.ObjectLengthName);
```
cincuranet: I think this isn't correct. It might collide with other entities/objects or with other objects in database. Best we can do is to throw an exception.

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Utilities/FbSettings.cs
```diff
@@ -0,0 +1,81 @@
...
+
+namespace FirebirdSql.EntityFrameworkCore.Firebird.Utilities
+{
+
+	public class FbSettings
+	{
+
+		public Version ServerVersion;
+		public bool IsSupportIdentityIncrement => ServerVersion.Major >= 3;
+		public int ObjectLengthName => ServerVersion.Major == 3  ? 31 : 64;
```
cincuranet: On FB3 it's 63 max. And it's configurable. So you really don't know until you try...

ralmsdeveloper:
Thanks Cincura, I was looking at this, I'll redo it!

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Storage/Internal/FbSqlGenerationHelper.cs
ralmsdeveloper: Truth! I'll do it!

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Extensions/FbStringExtension.cs
ralmsdeveloper: I know this, the question is to keep making several substring.

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Internal/FbOptions.cs
ralmsdeveloper: I do not need the new!
You can not find "FindExtension"

And the code gets a lot better!

cincuranet: What?

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Extensions/FbStringExtension.cs
cincuranet: Why? You can call `Substring(Math.Min)` and that's it.

ralmsdeveloper: I'll be using math in every way.
More if you like it that way. I'll do.
I want those comments for me to improve.

So you think this is better?
identifier.Substring (0, Math.Min (identifier.Length, _options.Settings.ObjectLengthName))

cincuranet: Could be. But I think because of [this](https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/pull/72#discussion_r138803619) the discussion is pointless.

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Internal/FbOptions.cs
ralmsdeveloper: I would have to see this in Microsoft.EntityFrameworkCore because it does not return.

Some others use it like this (Npgsql, Mysql (Oracle) ...

And I have noticed that it is indeed the best form!

I tested it and it just does not work in its way.

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Internal/FbOptions.cs
ralmsdeveloper: Would it look better?

public void Initialize(IDbContextOptions options)
{
    var fbOptions = GetOptions(options); 
    _fbSettings = new Lazy<FbSettings>(() => fbOptions.Connection != null
    ? new FbSettings().GetSettings(fbOptions.Connection)
    : new FbSettings().GetSettings(fbOptions.ConnectionString)); 
}

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Migrations/FbMigrationsSqlGenerator.cs
ralmsdeveloper: Give me a suggestion / example. Because only then will I do the others the way you need to!

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Migrations/FbMigrationsSqlGenerator.cs
ralmsdeveloper: Give me a suggestion / example. Because only then will I do the others the way you need to!

cincuranet: https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/pull/72/files#diff-56b8916bb1616666ceb41c2927e4eaa1R41

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Migrations/FbMigrationsSqlGenerator.cs
cincuranet: https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/pull/72/files#diff-56b8916bb1616666ceb41c2927e4eaa1R41

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Internal/FbOptions.cs
ralmsdeveloper: I will test again. I do not know why? More with me I tested a lot before applying this!

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Storage/Internal/FbSqlGenerationHelper.cs
zabulus: In our code we made hashing of names that don't fit in 31 symbols restrictions. Looks ugly at database view but without collisions (though not guaranteed, but still). I'd prefer to have my custom implementation for this, so I could configure Escaping from our side.

zabulus:
IMO, it's better to add some tests regarding all implementation. Just my 50 cents.

cincuranet:
Agreed. But I "finished" the stuff only two days ago. It's nowhere near ready for show time. And there's only so much free time I can spend on it.

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Internal/FbOptions.cs
ralmsdeveloper: **He calls here in every way!**

https://github.com/aspnet/EntityFrameworkCore/blob/dev/src/EFCore.SqlServer/Update/Internal/SqlServerModificationCommandBatchFactory.cs#L55

ralmsdeveloper: This is redundant.
The best way is what I suggested!
is my humble suggestion

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Internal/FbOptions.cs
cincuranet: Yes.

cincuranet: Also your `fbOptions` can be `null` and you should probably handle that.

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Internal/FbOptions.cs
cincuranet: No it's not the best way. The `FindExtension` is [using hastable lookup internally](https://github.com/aspnet/EntityFrameworkCore/blob/dev/src/EFCore/DbContextOptions.cs#L48), hence it's more efficient. You're iterating through the enumerator.

I suppose the `SqlServerModificationCommandBatchFactory.cs` wasn't updated yet to use the method available.

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Internal/FbOptions.cs
ralmsdeveloper: I will correct the formatting!

Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Internal/FbOptions.cs
ralmsdeveloper: I figured out why it does not work!
I needed a CAST in the "FbDbContextOptionsExtensions.cs" file

**From:** var extension = GetOrCreateExtension (optionsBuilder) .WithConnectionString (connectionString)

**To:** var extension = (FbOptionsExtension) GetOrCreateExtension (optionsBuilder) .WithConnectionString (connectionString);

ralmsdeveloper: Everything worked! :) rss

ralmsdeveloper:
Cincura I made the necessary corrections.
As for the question "Find Extension", I made a CAST based on this example and it worked! But without this CAST we will never succeed!

https://github.com/aspnet/EntityFrameworkCore/blob/dev/src/EFCore.SqlServer/Extensions/SqlServerDbContextOptionsExtensions.cs#L35

zabulus:
Tests could be reused from EF.Core repository, just copy and replace Fixture base methods. Adoptions will be needed of course, still better than writing from scratch.

cincuranet:
Why you closed it?

ralmsdeveloper:
Jiri, I closed myself so I could order more before sending, so you'll spend less time reviewing and requesting revisions.

I'll adjust the fork and build on what you've done by trying the best you can.

And sending it slowly!

ralmsdeveloper:
The intention is to leave in conformity with what you like.

PR #73, opened 2017-10-10
Using absolute time for connection lifetime checking (DNET-787)

cincuranet:


cincuranet:
Does this makes sense? @zabulus @mrotteveel @gerdus @ralmsdeveloper

zabulus:
Not much, specially for Environment.TickCount, as from https://msdn.microsoft.com/en-us/library/system.environment.tickcount(v=vs.110).aspx:
```Because the value of the TickCount property value is a 32-bit signed integer, if the system runs continuously, TickCount will increment from zero to Int32.MaxValue for approximately 24.9 days, then jump to Int32.MinValue, which is a negative number, then increment back to zero during the next 24.9 days.```

What are you trying to achieve and what was the problem with DateTimeOffset?

cincuranet:
The overflow is handled [here](https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/pull/73/files#diff-78a68475f171c75864fd91db05f9e669R168). 

Check the [DNET-787](http://tracker.firebirdsql.org/browse/DNET-787) for the problem. I personally don't see it as a huge problem. But I gave it a shot anyway.

gerdus:
The IsAlive logic is a bit dense to understand.

maybe use System.Diagnostics.Stopwatch.GetTimestamp() instead?
according to (https://msdn.microsoft.com/en-us/library/windows/desktop/dn553408(v=vs.85).aspx) it only rolls over every 100 years

cincuranet:
The problem with `System.Diagnostics.Stopwatch.GetTimestamp` is that in case high-resolution performance counter is not available it falls back to DateTime. :( Not sure I want to risk some hidden issues.

zabulus:
```Check the DNET-787 for the problem. I personally don't see it as a huge problem. But I gave it a shot anyway.```
In this case, IMHO, change is OK.

gerdus:
Think the change as-is is good and more correct than it was previously.

FYI Stopwatch.GetTimestamp with high resolution looks pretty well supported (http://aakinshin.net/blog/post/stopwatch) Mono,Unix,Mac,Windows XP+

ralmsdeveloper:
If we consider an optimal connection after I / O is successful (10,000 ticks = 1 ms) Approximately.
The resolution of the timer is usually in the range of 10 to 16 milliseconds.

Environment.TickCount is based on the GetTickCount () function. E is in milliseconds, but the actual precision is about 15.6 ms. In this way we can not measure very small intervals of time, because we would certainly receive 0.

I think the idea of ​​using System.Diagnostics. Stopwatch better.

But this is my opinion. and I also think the work you did is good.

cincuranet:
Considering the connection lifetime is in seconds the timer resolution
doesn't matter.

ralmsdeveloper:
Okay @cincuranet , I just wanted to emphasize the use of Stopwatch.

I think the person who will use something of this magnitude will have to treat it as well.

cincuranet:
You're right @gerdus, the support on non-exotic platforms is indeed good. I decided to [write it with `Stopwatch`](https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/compare/dnet787...dnet787v2) just to see how it feels and run some tests.

I've also found this:

img

Although it probably does not matter that much for the connection pooling.

As I was writing it I realized I can maybe the value from `Environment.TickCount` convert to `long` right away and make it non-negative. Then the `IsAlive` will be simpler. Then just hope nobody will have a connection in pool for longer than ~49 days with longer lifetime. :)

mrotteveel:
Looks good to me

PR #75, opened 2017-12-15
fixed IsAlive calculation in FbConnectionPoolManager.Pool.IsAlive

realic:
Hey Jiri
I found an error in version 5.12.0. I believe it was introduced when fixing DNET-787. 

Please advise if the unittest is not implementet in a way you like.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ packages
 DDEX/installer/out/
 DDEX/installer/in/
 Provider/deploy/out/
+.idea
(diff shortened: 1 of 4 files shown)


cincuranet:
Fix the PR to not have unnecessary changes, please.

realic:
Do you want me to extract the IsAlive and GetTicks to a seperate publice testable class?

cincuranet:
Yes. I would put into `Common` something like `ConnectionLifetimeHelper` (internal), with `IsAlive` method. And I would test this. The `GetTicks` is not needed.

realic:
The requested changes have been implemented, if you haven't already received a notification about it.

cincuranet:
Merged by 138a347f0f3c19a0ff9b39d3cdd94a002d3b1b8c.

PR #76, opened 2018-04-18
Fix for Xamarin Android and .NET Standard 2.0

dlichterman:
This fix will make the .NET Standard 2.0 version work with Xamarin.Android. Xamarin.Android doesn't support Microsoft.Extensions.PlatformAbstractions which seems to be something that is not being supported any more with some new features in .NET Standard 2.0: https://github.com/aspnet/Announcements/issues/237

This will leave 1.6 using Microsoft.Extensions.PlatformAbstractions, but for the compile option NETSTANDARD2_0 it will switch to System.AppContext.BaseDirectory.


tessarolli:
Waiting for approval..

cincuranet:
Looks OK to me in general. I'll merge it later and it will go into some 6.x version.

cincuranet:
Little small cleanup and merged by c32f6035. Thanks for the contribution.

PR #77, opened 2018-07-25
MigrationHistory hex literal limit workaround

SingThatSong:
An attempt to take on http://tracker.firebirdsql.org/browse/DNET-769

What I'm doing is splitting one big INSERT when migrating EF into smaller INSERT and number of UPDATE queries.

There may be a way to do this task in more elegant way, so it requires review. Also, not sure how to test it - I'm using it in my project, works fine for now.

### DIFF
--- a/Provider/src/EntityFramework.Firebird/FbMigrationSqlGenerator.cs
+++ b/Provider/src/EntityFramework.Firebird/FbMigrationSqlGenerator.cs
@@ -17,10 +17,12 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.Data.Common;
 using System.Data.Entity;
 using System.Data.Entity.Core.Common;
 using System.Data.Entity.Core.Common.CommandTrees;
+using System.Data.Entity.Core.Common.CommandTrees.ExpressionBuilder;
 using System.Data.Entity.Core.Metadata.Edm;
 using System.Data.Entity.Infrastructure.DependencyResolution;
 using System.Data.Entity.Migrations.Model;
@@ -30,6 +32,7 @@ using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
 using EntityFramework.Firebird.SqlGen;
+using FirebirdSql.Data.Common;
 
 namespace EntityFramework.Firebird
 {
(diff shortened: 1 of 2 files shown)


Provider/src/EntityFramework.Firebird/FbMigrationSqlGenerator.cs
```diff
@@ -447,11 +450,101 @@ protected virtual IEnumerable<MigrationStatement> Generate(HistoryOperation oper
...
+										dbInsert.Returning);
+
+									writer.Write(DmlSqlGenerator.GenerateUpdateSql(updateCommandTree, out _, generateParameters: false));
+
+									// Since we couldn't concat before, replacing query as string
+									// Replacing SET Model = 'data'
+									//		with SET Model = Model || 'data'
+									// Model being first is important, since these are parts of single value
+									var statement = writer.ToString();
+									var newStatement = statement.Replace($"SET \"{modelProperty.Property.Name}\" = ",
```
cincuranet: Hmm. Wondering whether this could be made without _string replace_.

SingThatSong: Yeah, thats kinda ugly.

My first thought was to do it like
`((DbExpression)modelProperty.Property.Name).Concat(dataChunk.ToArray())` at [row 521](https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/pull/77/files#diff-dbd57665224c10e3d0cccd240e4ffb9dR521) where it should be, but it returns `DbFunctionExpression`, which is [not supported](https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/blob/48f30748e682a45e83a3041b4e97d831e7dd2e84/Provider/src/EntityFramework.Firebird/SqlGen/ExpressionTranslator.cs#L119) by `ExpressionTranslator` and I have no idea how to implement it properly for all types of functions.

cincuranet: Yeah. One implementation is [here](https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/blob/48f30748e682a45e83a3041b4e97d831e7dd2e84/Provider/src/EntityFramework.Firebird/SqlGen/SqlGenerator.cs#L2342), because for selects it's needed. Maybe the core could be extracted into some helper and used in both places. 🤔 

Or maybe the string replace version is just good enough. I'm not sure. 😊

PR #78, opened 2018-07-26
 Add explicit command logging switch 

gerdus:
In my current project (.net core command line app) I just can't seem to disable the tracing with the version from nuget. With the tracing enabled the ETL program I am debugging is extremely slow in VS 2017.
It is always a pain to remember how to disable it. Thus this tiny change.
So I can easily disable command logging with:
`
FirebirdSql.Data.FirebirdClient.FbCommandOptions.EnableCommandLogging = False
`
In my opinion it should default to disabled.

Thank You

### DIFF
--- a/Provider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbCommand.cs
+++ b/Provider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbCommand.cs
@@ -32,6 +32,10 @@ using FirebirdSql.Data.Common;
 
 namespace FirebirdSql.Data.FirebirdClient
 {
+	public static class FbCommandOptions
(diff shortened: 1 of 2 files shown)


cincuranet:
I don't think this is a good change. A) it's non-standard, thus hard to discover and B) gives another option to turn off/on the logging, which might make it difficult to manage.

I have a [DNET-782](http://tracker.firebirdsql.org/browse/DNET-782) in the tracker, that would "solve" this. Maybe I can increase the priority and prepare it for next version. ;)

gerdus:
I agree Microsoft.Extensions.Logging sounds like a better long term option. No need to increase the priority on my part, I'll just keep using my local version til implemented.

thank you. 👍

PR #84, opened 2019-06-21
 refactor XdrStream to reduce copying data between buffers

dant02:
goal is to speed up reading / writing from / to network stream by reducing copies between byte arrays

### DIFF
--- a/Provider/src/FirebirdSql.Data.FirebirdClient/Client/Managed/ByteQueue.cs
+++ b/Provider/src/FirebirdSql.Data.FirebirdClient/Client/Managed/ByteQueue.cs
@@ -13,7 +13,7 @@
  *    All Rights Reserved.
  */
 
-//$Authors = Daniel Trubač
+//$Authors = Daniel Trubac
 
 using System;
 using System.IO;
@@ -28,15 +28,10 @@ namespace FirebirdSql.Data.Client.Managed
 		protected int readIndex = 0;
 		protected int writeIndex = 0;
 
-		public ByteQueue(string name)
+		public ByteQueue(string name, int size)
 		{
 			this.Name = name;
-			this.buffer = new byte[allocatedSize];
-		}
-
-		public ByteQueue(string name, int size) : this(name)
-		{
-			allocatedSize = size;
+			this.allocatedSize = size;
 			this.buffer = new byte[allocatedSize];
 		}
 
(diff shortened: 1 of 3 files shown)


cincuranet:
Hi, I'm currently working on something else. I'll review this later.

cincuranet:
I tried running simple benchmark that fetches 200k `BIGINT`s and 200k `VARCHAR`s with this PR and with 6.6.0. This the result:

* 6.6.0

| Method |     Mean |    Error |   StdDev |      Gen 0 | Gen 1 | Gen 2 | Allocated |
|------- |---------:|---------:|---------:|-----------:|------:|------:|----------:|
|  Int64 | 303.9 ms | 2.803 ms | 2.622 ms | 22500.0000 |     - |     - |  68.08 MB |
| String | 371.3 ms | 6.846 ms | 6.404 ms | 28000.0000 |     - |     - |   86.4 MB |

* PR

| Method |     Mean |    Error |   StdDev |      Gen 0 | Gen 1 | Gen 2 | Allocated |
|------- |---------:|---------:|---------:|-----------:|------:|------:|----------:|
|  Int64 | 303.1 ms | 3.811 ms | 3.565 ms | 22500.0000 |     - |     - |  67.85 MB |
| String | 370.9 ms | 5.649 ms | 5.008 ms | 28000.0000 |     - |     - |  86.16 MB |

The times are virtually the same, Gen 0 collection are exactly the same. Only in terms of allocated memory a small improvement. 

Do you see different numbers?

dant02:
Hi,
please see this https://github.com/dant02/snippets/blob/master/dotnet/app.console/Program.cs

so far i have noticed that main difference is in transfer of OP codes, (not sure if its correct term or if i understand correctly how firebird protocol works), so when there are actual data it makes little difference,

but when there are NULL values, firebird protocol upkeep gets in a way

for the example above without PR most work done is in .net process, with PR most work done is in FB process, difference on my pc is 37 seconds vs 28 seconds

cincuranet:
This is what I get with your code:

* 6.6.0

| Method |    Mean |   Error |   StdDev |        Gen 0 |     Gen 1 | Gen 2 | Allocated |
|------- |--------:|--------:|---------:|-------------:|----------:|------:|----------:|
|  Dummy | 19.40 s | 4.697 s | 0.2575 s | 2097000.0000 | 1000.0000 |     - |   6.14 GB |

* PR

| Method |    Mean |   Error |   StdDev |        Gen 0 |     Gen 1 | Gen 2 | Allocated |
|------- |--------:|--------:|---------:|-------------:|----------:|------:|----------:|
|  Dummy | 20.60 s | 2.978 s | 0.1632 s | 2094000.0000 | 2000.0000 |     - |   6.14 GB |

dant02:
Hi, would you please point me how i can measure the code same way

cincuranet:
I'm using [BenchmarkDotNet](https://benchmarkdotnet.org/) with `MemoryDiagnoser`. Nothing specific.

dant02:
Ok, im closing this PR since its in conflict with new version, thank you

cincuranet:
You can always rebase, but sure.

PR #86, opened 2019-09-08
Fix NotImplementedException when using string or byte array fields.

siegfriedpammer:
This pull request contains all the changes that were necessary to make our application to work with EF Core.

The `NotImplementedException`s we were running into were thrown [here](https://github.com/aspnet/EntityFrameworkCore/blob/release/2.2/src/EFCore.Relational/Storage/RelationalTypeMapping.cs#L446)

Example exception text:
`System.NotImplementedException:` 'The `FbByteArrayTypeMapping` does not support value conversions. Support for value conversions typically requires changes in the database provider.'

Let me know whether this is the correct fix.

### DIFF
--- a/Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Storage/Internal/FbBoolTypeMapping.cs
+++ b/Provider/src/FirebirdSql.EntityFrameworkCore.Firebird/Storage/Internal/FbBoolTypeMapping.cs
@@ -28,9 +28,18 @@ namespace FirebirdSql.EntityFrameworkCore.Firebird.Storage.Internal
 			: base("BOOLEAN", System.Data.DbType.Boolean)
 		{ }
 
+		private FbBoolTypeMapping(RelationalTypeMappingParameters parameters)
+			: base(parameters)
+		{ }
+
 		protected override string GenerateNonNullSqlLiteral(object value)
 		{
 			return (bool)value ? TrueLiteral : FalseLiteral;
+		}
+
+		protected override RelationalTypeMapping Clone(RelationalTypeMappingParameters parameters)
+		{
+			return new FbBoolTypeMapping(parameters);
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


cincuranet:
Hi, thanks for the PR. Given the changes are unrelated, please split it into two PRs.

cincuranet:
Any reason why only `FbByteArrayTypeMapping` and `FbStringTypeMapping` is changed and not the others?

siegfriedpammer:
Sorry for taking so long to get back to this... I am rather busy currently.

> Any reason why only `FbByteArrayTypeMapping` and `FbStringTypeMapping` is changed and not the others?

Only those were failing in our specific usage case. Of course I will add the other mappings as well, if you think it is necessary.

Do I see correctly, that there were no build failures, just cancellations?

siegfriedpammer:
The only type mapping that is "non-trivial" to clone are the `FbDateTimeTypeMapping` and `FbStringTypeMapping`. I will have to find a way to include the `FbDbType` field in the `RelationalTypeMappingParameters` provided by EF.

Any suggestions/ideas how we could accomplish this? Subclassing `RelationalTypeMappingParameters` seems to be an awful idea.

cincuranet:
Do you have some code where the exception happens? I'd like to include it into the test suite.

siegfriedpammer:
I'll have to dive in again and try to produce a minimal sample. Not sure, when I will have time again. Hopefully I will be able to get to this in a few days.

cincuranet:
OK. Great.

siegfriedpammer:
[FBEFCorePR86Repro.zip](https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/files/3824484/FBEFCorePR86Repro.zip)

Simply unzip the attached solution, open it in Visual Studio and run the code. After a successful restore of the embedded database it should crash with a NotImplementedException.

DDL and GBK included in the data folder.

I hope this is sufficient for building a unit test.

Thank you again for your patience!

cincuranet:
BTW to answer how to handle the passing of `FbDbType`, I needed the `Clone` myself now when working on EF Core 3.1 and I simply passed it as a value for the `ctor`, see [here](https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/commit/ac48a2d2dde66d3b0bbc416c3660e4c33fd63283#diff-a79c5c58eb7a020cf338c6a16d013f83).

I'll check your example code as well later, when provider and 3.1 is done-ish. Maybe it will be solved as together with the 3.1 work as a side effect.

cincuranet:
OK, this has been solved as part of EF Core 3.1 implementation. Your, BTW *very* good, test bench runs fine. I'll be releasing the final version today/tomorrow.

PR #88, opened 2019-10-19
reduce copies of byte arrays when reading blob

dant02:


Provider/src/FirebirdSql.Data.FirebirdClient/Client/Managed/Version10/GdsBlob.cs
```diff
@@ -119,40 +119,28 @@ protected override byte[] GetSegment()
 
 				RblRemoveValue(IscCodes.RBL_segment);
 				if (response.ObjectHandle == 1)
-				{
 					RblAddValue(IscCodes.RBL_segment);
-				}
+
```
cincuranet: Please keep the brackets (across the whole PR).

Provider/src/FirebirdSql.Data.FirebirdClient/Client/Native/FesBlob.cs
```diff
@@ -160,7 +160,7 @@ protected override byte[] GetSegment()
 					}
 				}
 
-				return segment.ToArray();
+				segment.WriteTo(ms);
```
cincuranet: Why not to write directly to `ms` above?

Provider/src/FirebirdSql.Data.FirebirdClient/Common/BlobBase.cs
```diff
@@ -130,7 +127,7 @@ public void Write(byte[] buffer, int index, int count)
 
 		protected abstract void Create();
 		protected abstract void Open();
-		protected abstract byte[] GetSegment();
+		protected abstract void GetSegment(MemoryStream ms);
```
cincuranet: Maybe just use plain `Stream`, because nothing requires the parameter to be `MemoryStream` and I'd call the parameter `stream`.

Provider/src/FirebirdSql.Data.FirebirdClient/Client/Managed/Version10/GdsBlob.cs
dant02: Added back

Provider/src/FirebirdSql.Data.FirebirdClient/Client/Native/FesBlob.cs
dant02: That i did not notice, refactored in following commit

Provider/src/FirebirdSql.Data.FirebirdClient/Common/BlobBase.cs
dant02: Yes your right, its better that way, refactored in following commit

cincuranet:
Thanks.

PR #90, opened 2019-11-08
Test broken connection handling with connectionpooling enabled

trynn:
hey,
test will fail until http://tracker.firebirdsql.org/browse/DNET-917 is also solved, because simulation via mon$attachments will not work. Maybe you have a better idea how to simulate it.


trynn:
closed in favor of PR91

PR #91, opened 2019-11-12
Busy list in pool handling in case of an error (DNET-916).

cincuranet:


cincuranet:
PTAL @trynn. My spin on DNET-916.

trynn:
tested with unittest from PR90 and manually. should work too.
whatever works is fine for me.

i would really like to see unittests for the pooling, since pretty time consuming testing new versions always for the same issue. if you can find some time. my disconnect approach from PR90 will only work after fixing DNET-917.

cincuranet:
Do PR with only the test from #90 and we can start from it.

PR #92, opened 2019-11-13
Configurable TransactionBehaviors for specific IsolationLevel

cmayer0087:
With this you can alter the default behavior for the IsolationLevels.
We need this because we use an ORM-Wrapper. So we can't use the constructor with FbTransactionOptions parameter and our code relies on the Wait used back then with FirebirdClient 2.

For example:
```
<?xml version="1.0">
<configuration>
  <configSections>
    <section name="firebirdSettings" type="System.Configuration.NameValueSectionHandler"/>
  </configSections>
  <firebirdSettings>
    <add key="TransactionBehavior.Chaos" value="Wait, ReadCommitted, NoRecVersion"/>
    <add key="TransactionBehavior.ReadCommitted" value="Wait, ReadCommitted, NoRecVersion"/>
  </firebirdSettings>
</configuration>
```

### DIFF
--- a/Provider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbTransaction.cs
+++ b/Provider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbTransaction.cs
@@ -18,6 +18,8 @@
 using System;
 using System.Data;
 using System.Data.Common;
+using System.Configuration;
+using System.Collections.Specialized;
 
 using FirebirdSql.Data.Common;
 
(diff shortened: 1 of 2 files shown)


cincuranet:
I don't think this is going to be merged. Do you know any other ADO.NET provider (SqlClient, Npgsql, …) having same/similar feature?

Also why would provider supply lack of feature - custom transaction usage - in some ORM? EF6, EFCore, Dapper, … support it. 

Also in Core the app/web.config is generally not used, so something with better fit would be worth thinking about.

cmayer0087:
There is a difference behavior between SqlClient and FirebirdClient.

In this example: SqlClient will wait on line 20, which FirebirdClient don't in a similar code.
```
1            using(SqlConnection con1 = new SqlConnection("Server=.\\SQL2016;Database=test;Trusted_Connection=True"))
2            using(SqlConnection con2 = new SqlConnection("Server=.\\SQL2016;Database=test;Trusted_Connection=True"))
3            {
4                con1.Open();
5                
6                SqlTransaction t1 = con1.BeginTransaction(System.Data.IsolationLevel.ReadCommitted);
7                
8                SqlCommand cmd = con1.CreateCommand();
9                cmd.Transaction = t1;
10               cmd.CommandText = "UPDATE test SET Value = 'TestValue1' WHERE ID = 1";
11               cmd.ExecuteNonQuery();
12
13               con2.Open();
14                
15               SqlTransaction t2 = con2.BeginTransaction(System.Data.IsolationLevel.ReadCommitted);
16
17               cmd = con2.CreateCommand();
18               cmd.Transaction = t2;
19               cmd.CommandText = "UPDATE test SET VALUE ='TestValue2' WHERE ID = 1";
20               cmd.ExecuteNonQuery();
21           }
```
Older versions of FirebirdClient had the same behavior as SqlClient.
We tracked that down to changed code in BuildTpb and wanted a way to restore the old behavior.

How should that be implemented? 
Would be an additional connectionstring parameter - like "WaitForLock" - a reasonable solution?

BTW: There is a way to pass the isolationlevel to our ORM like in IDbConnection.BeginTransaction but no way to pass custom FbTransactionOptions.

cincuranet:
> There is a difference behavior between SqlClient and FirebirdClient.
> In this example: SqlClient will wait on line 20, which FirebirdClient don't in a similar code.

Yeah, that's not an argument. It's a difference in how transactions behave in the engines, not how provider does something. Same as saying one has `SELECT FIRST ...`, while other `SELECT TOP ...`. 

> Would be an additional connectionstring parameter - like "WaitForLock" - a reasonable solution?

Not only would that be a wrong name, it would make sense only for some cases and duplicate what's already possible in code. So no, that's not a path to go.

> BTW: There is a way to pass the isolationlevel to our ORM like in IDbConnection.BeginTransaction but no way to pass custom FbTransactionOptions.

It's not about passing isolation level, but about passing `DbTransaction` object. I.e. [this in EF6](https://docs.microsoft.com/en-us/dotnet/api/system.data.entity.database.usetransaction?view=entity-framework-6.2.0).

And I still think, this is _correct_ way to do it. Not hacking it in provider. But I'm open to ideas and arguments (reasonable).

cmayer0087:
Ok, i see.

PR #94, opened 2020-03-16
added settings for C# whitespace handling

Kaupisch-IT:
Added Visual Studio .editorconfig settings for C# whitespace handling
(so VS does not mess up e.g. spaces after commas or in for statements)


Kaupisch-IT:
For spaces around `==` (resp. binary operators in general) change the corresponding line to
`csharp_space_around_binary_operators = true`

cincuranet:
Update the PR, please.

Kaupisch-IT:
I can't, the repo is gone. :-P
However, the settings are only educated guesses that at least don't mess up whitespace in existing code. Exporting your personal settings (https://docs.microsoft.com/en-us/visualstudio/ide/code-styles-and-code-cleanup?view=vs-2019) should be more reliable I think.

PR #96, opened 2020-04-02
add connection timeout to connection to consider "connection timeout="

litti:
add connection timeout to connection to consider "connection timeout=" parameter in connection string


cincuranet:
Sorry for taking time to respond. The reason is, I knew I'm (was) working on full async support and with that came also the `CancellationToken` related stuff. 

With that in place the async path supports cancellation via the token. I just need to think about a good way to combine explicitly provided `CancellationToken` and the "implicit" value from connection string.

Not sure though about the sync path. Is it worth it (and probably switching between blocking and non-blocking socket together with `Socket.Select` would be a better way to go)?

Finally, more as a note for myself, the `TimeoutException` is worth considering as an exception for the timeout.

cincuranet:
BTW [this](https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/commit/c97c4046) commit added the handling (on top of 8.0 code). At the moment handling also sync code path seems to be waste of effort.

Do you see anything else missing?

PR #99, opened 2020-04-28
Implemented Scaffolding, Added Blob Mapping from Db to Model

pgiacomo69:
Implemented Scaffolding, Added Blob Mapping from Db to Model


cincuranet:
Can you add/extract the blob mapping into separate PR and maybe use `BLOB SUB_TYPE BINARY` instead of number?

pgiacomo69:
ok

> Il giorno 7 giu 2020, alle ore 20:04, Jiri Cincura ↹ <notifications@github.com> ha scritto:
> 
> Can you add/extract the blob mapping into separate PR and maybe use BLOB SUB_TYPE BINARY instead of number?

cincuranet:
You don't have to close this one. Just the one change into separate PR, rest can stay.

pgiacomo69:
Done,
sorry but I’m not a git expert, I preferred to pull down the entire fork, I will push other modifications one by one.I think  I’ve to study how to work with branches.

PR #100, opened 2020-06-07
Added Blob Mapping from Db to Model

pgiacomo69:


cincuranet:
Thanks.

pgiacomo69:
Nothing :)

I’ve three other PR to do, but as I said I don’t know how to keep them separate, in which order should I post PRs?:

1) FbMigrationsSqlGenerator.cs, 213

	protected override void Generate(DropIndexOperation operation, IModel model, MigrationCommandListBuilder builder, bool terminate = true)
		{
			builder.Append("DROP ");
			IndexTraits(operation, model, builder);
			builder.Append("INDEX ");
			builder.Append(Dependencies.SqlGenerationHelper.DelimitIdentifier(operation.Name));
			if (terminate)
				TerminateStatement(builder);
		}

Since on add-migration the default Method  is unable to  drop the indexes;

2) FbUpdateSqlGenerator.cs, 203
	public override string GenerateNextSequenceValueOperation(string name, string schema)
		{
			var builder = new StringBuilder();
			builder.Append("SELECT NEXT VALUE FOR ");
			builder.Append(SqlGenerationHelper.DelimitIdentifier(name));
			builder.Append(" FROM RDB$DATABASE");
			return builder.ToString();
		}
Without this override you are unable to get the next value for a sequence;

3) FbDatabaseModelFactory.cs

Scaffolding Implemantation, as you saw on my old PR, reformatted and corrected as you requested.

Thank You,
Giacomo

cincuranet:
> I’ve three other PR to do, but as I said I don’t know how to keep them separate, in which order should I post PRs?:

Create 3 separate branches and create PRs for each - after committing GitHub will show you link in your repo's page to create that PR.

PR #107, opened 2021-01-20
Fix DNET-998

lweberprb:
Fixes the issue that the connection pool was seemingly getting full very quickly, after connection to the database has been lost. The created connections were put in the busy list, as expected. If, however, the created connection could not be connected to the database, it would remain in the collection of busy connection

### DIFF
--- a/Provider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbConnectionPoolManager.cs
+++ b/Provider/src/FirebirdSql.Data.FirebirdClient/FirebirdClient/FbConnectionPoolManager.cs
@@ -91,7 +91,15 @@ namespace FirebirdSql.Data.FirebirdClient
 				}
 				if (createdNew)
 				{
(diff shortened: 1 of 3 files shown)


lweberprb:
With the latest changes on the "fbconn-async" branch, the problematic section of code is removed, therefore this fix is probably only relevant for the current state of the master branch. 
Is there any possbility, that this pull request makes its way into the nuget-package as a hotfix for the current problem?

lweberprb:
I'have dug a little deeper into the fbconn-async branch: The issue still persists there, it has only moved to a different section of code. Now FbConnection.cs/OpenImpl connects the Connection after getting it from the Connection pool, but does not return it, if the initial connect has failed. I'll be opening a new Pull-Request for that.

cincuranet:
Thanks fixed in 2290ec05a3493ae61ba67d19f245518315f63d51, using basically the same code (and also added test for this case).

PR #108, opened 2021-01-23
Fix DNET-998 (async)

lweberprb:
Same problem as #107. Fix moved to different code section taking into account the latest changes for an fully async implementation of the .NET-Provider


cincuranet:
Thanks fixed in 2290ec05a3493ae61ba67d19f245518315f63d51, using basically the same code (and also added test for this case).

PR #111, opened 2021-04-08
Fix scaffolding indices

lweberprb:
As per https://groups.google.com/g/firebird-net-provider/c/YyMegW_RwvQ/m/Y9DarC5eAgAJ


cincuranet:
Thanks!

PR #112, opened 2021-05-05
Update the name of charset to match the exact name of the charset

omarbenzekri:
old value Windows1252
new value WIN1252


cincuranet:
Explain why would `FbCharset` need to match anything?

omarbenzekri:
Hi ,sorry to be latei had a problem with this line:
  constring.Charset = FbCharset.WIN1256.ToString();--------->before it was constring.Charset = FbCharset. Windows1252.ToString();
constring is FbConnectionStringBuilder.

if the returned string doesn't match the value in FirebirdSql.Data.Common.charset.cs when the finction GetCharset(string charsetName) is called we get an exception.
i hope i was clear.best regardes    Le mercredi 5 mai 2021, 12:34:43 UTC+1, Jiri Cincura ↹ ***@***.***> a écrit :  
 
 


Explain why would FbCharset need to match anything?

cincuranet:
The connection string charset has no relation to to `FbCharset` enum. That's why it's also `string` in `FbConnectionStringBuilder`. 

Although it could be changed, it's going to be a breaking change and hence requiring new major version. And also thinking about the pros and cons of doing that.

For now I'm going to close this (also because the changes you did are heavily backward incompatible). If you feel like driving this forward, I suggest opening a discussion in mailing list and eventually later providing full implementation.

omarbenzekri:
ok thank you for your time    Le samedi 8 mai 2021, 09:48:18 UTC+1, Jiri Cincura ↹ ***@***.***> a écrit :  
 
 


Closed #112.

PR #114, opened 2023-07-10
Tests improvements

fdcastel:
This PR changes the folowing:

1. **Adds missing packages for NUnit-based tests.**
    When running tests in Visual Studio 2022 only xUnit-based tests work. 
    
    Sources: [here](https://stackoverflow.com/a/74467458/33244) and [here](https://stackoverflow.com/a/67433575/33244).

2. **Make it easier for new users to run tests.**
    Uses `$env:TEMP\fb_tests\$FirebirdSelection\` instead of `I:\Downloads\fb_tests`. 

    This hardcoded default, together with a silent exception being throw (see item 4. below) made it very hard for me to understand what was happening until I debug the script.

3. **Makes the download of Firebird binaries much faster.**
    Due to a [bad implementation](https://stackoverflow.com/a/43477248/33244) of progress in `Invoke-RestMethod`.

4. **Makes the script more resilient to aborts.**
    Hitting `Ctrl-C` during a test run often left Firebird process running and the current directory (`$PWD`) changed. 
    
    The root cause being a call to `$MyInvocation.MyCommand.Name` throwing a silent exception and not running the actual cleanup code.


fdcastel:
Sorry. I don't understand why GitHub is showing `13` commits. I based my branch/pr on [this commit](https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/pull/114/commits/3f2894448a158fd71388052e763daf5ceebe89fd).

This PR should have only `5` commits.

cincuranet:
Please create separate PRs for each change. Also you need to start from master branch.

fdcastel:
Understood. Sorry it was late night 😉. Fixing...

On the other hand... Separate **PRs** for each change? Isn't this a bit excessive? It's all related. 

Didn't you mean separate **COMMITS**? (which was what already done?)

cincuranet:
Nope. I meant separate PRs.

PR #115, opened 2023-07-10
Adds Microsoft.NET.Test.Sdk and NUnit3TestAdapter to NUnit projects.

fdcastel:
1. **Adds missing packages for NUnit-based tests.**
    When running tests in Visual Studio 2022 only xUnit-based tests work. 
    
    Sources: [here](https://stackoverflow.com/a/74467458/33244) and [here](https://stackoverflow.com/a/67433575/33244).

2. **Upgrades Microsoft.NET.Test.Sdk to version 17.6.3**


cincuranet:
Thanks.

PR #116, opened 2023-07-10
Uses $env:TEMP folder instead of hardcoded.

fdcastel:
1. **Make it easier for new users to run tests.**
    Uses `$env:TEMP\fb_tests\$FirebirdSelection\` instead of `I:\Downloads\fb_tests`. 

    This hardcoded default, together with a silent exception being throw (see [further PR](https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/pull/118)) made it very hard for me to understand what was happening until I debug the script.


cincuranet:
Thanks.

PR #117, opened 2023-07-10
Makes downloads with Invoke-RestMethod faster.

fdcastel:
1. **Makes the download of Firebird binaries much faster.**
    Due to a [bad implementation](https://stackoverflow.com/a/43477248/33244) of progress in `Invoke-RestMethod`.


cincuranet:
Thanks.

PR #121, opened 2023-07-12
Adds tasks for Invoke-Build.

fdcastel:
Port of `build.ps1` script to [`Invoke-Build`](https://github.com/nightroman/Invoke-Build/wiki).  Tell me what you think. If you did like I will port the other scripts.

To use it, just install `Invoke-Build` with

`dotnet tool install --global ib`

And call it with:

1.
    ```powershell
    ib
    ```
    equivalent to `.\build.ps1 -Configuration Debug`

2.
    ```powershell
    ib -Configuration Release
    ```
    equivalent to `.\build.ps1 -Configuration Release`

3. I also added a new `-VersionSuffix` parameter (optional). So it is now possible to build nuget packages with a specific version without changing `Directory.Build.props`. E.g.: 
    ```powershell
    ib -Configuration Release -VersionSuffix alpha2
    ```

### DIFF
--- a/FirebirdSql.Data.FirebirdClient.build.ps1
+++ b/FirebirdSql.Data.FirebirdClient.build.ps1
@@ -1,0 +1,62 @@
+#
+# FirebirdSql.Data.FirebirdClient Tasks for Invoke-Build
+#
+# Requires Invoke-Build -- https://github.com/nightroman/Invoke-Build
+#
+#   dotnet tool install --global ib
+#
+
+param(
+    $Configuration = 'Debug',
+    $VersionSuffix = $null
+)
+
+
+#
+# Globals
+#
+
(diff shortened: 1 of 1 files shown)


cincuranet:
What's the justification for this change?

fdcastel:
Not a change. The old scripts are still there. 

It's more like a sample. For you to see `Invoke-Build` in action. It's like `make`, but in Powershell.

There are several advantages in using it instead of ad-hoc build scripts. From memory, I could list:
- Incremental tasks
- Helper scripts (e.g. finds correct MSBuild)
- Parallel builds

But, of course, you need to use what is useful for you. As I said. I use it in all my projects.

cincuranet:
OK. Thanks. But no.

PR #125, opened 2023-09-10
Add RecordVersionStatistics flag on FbStatisticalFlags

gerdus:
Add missing option for analyze average record and version length.
Also xml document comments on the enum.


src/FirebirdSql.Data.TestsBase/FbTestsBase.cs
gerdus, line 1: Needed to add the port for service connection string builder since I run firebird on a non standard port.

cincuranet:
Looks good. Merging. Thanks.

PR #126, opened 2023-10-10
Support DESC-flag in migrations and scaffolding

lweberprb:
Add support for descending indices for EF-Core 7

As indices in Firebird are unidirectional, this PR looks only at the first flag in the list to determine whether the whole index is supposed to be descending. If this is not the behaviour that is being looked for, i am happy to modify my PR accordingly.

### DIFF
--- a/src/FirebirdSql.EntityFrameworkCore.Firebird/Migrations/FbMigrationsSqlGenerator.cs
+++ b/src/FirebirdSql.EntityFrameworkCore.Firebird/Migrations/FbMigrationsSqlGenerator.cs
@@ -201,6 +201,19 @@ public class FbMigrationsSqlGenerator : MigrationsSqlGenerator
 		{
 			builder.Append("UNIQUE ");
 		}
+
+		if (operation.IsDescending is not null && operation.IsDescending.Length > 0)
+		{
+			var isDescending = operation.IsDescending[0];
+			if (operation.IsDescending.Any(x => x != isDescending))
+				throw new NotSupportedException("Mixed order indices are not supported by Firebird.");
+
+			if (isDescending)
+			{
+				builder.Append("DESC ");
+			}
+		}
+
 		IndexTraits(operation, model, builder);
 		builder.Append("INDEX ");
 		builder.Append(Dependencies.SqlGenerationHelper.DelimitIdentifier(operation.Name));
--- a/src/FirebirdSql.EntityFrameworkCore.Firebird/Scaffolding/Internal/FbDatabaseModelFactory.cs
+++ b/src/FirebirdSql.EntityFrameworkCore.Firebird/Scaffolding/Internal/FbDatabaseModelFactory.cs
@@ -311,6 +311,7 @@ public class FbDatabaseModelFactory : DatabaseModelFactory
 		@"SELECT
                trim(I.rdb$index_name) as INDEX_NAME,
                COALESCE(I.rdb$unique_flag, 0) as IS_UNIQUE,
+               Coalesce(I.rdb$index_type, 0) = 1 as IS_DESC,
                list(trim(sg.RDB$FIELD_NAME)) as COLUMNS
               FROM
                RDB$INDICES i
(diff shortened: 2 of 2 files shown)


src/FirebirdSql.EntityFrameworkCore.Firebird/Migrations/FbMigrationsSqlGenerator.cs
```diff
@@ -201,6 +201,10 @@ protected override void Generate(CreateIndexOperation operation, IModel model, M
 		{
 			builder.Append("UNIQUE ");
 		}
+		if (operation.IsDescending?.Length > 0 && operation.IsDescending[0])
```
cincuranet, line 204: Probably worth throwing an exception when some columns are _asc_ and some _desc_.

src/FirebirdSql.EntityFrameworkCore.Firebird/Scaffolding/Internal/FbDatabaseModelFactory.cs
```diff
@@ -311,6 +311,11 @@ private void GetPrimaryKeys(DbConnection connection, IReadOnlyList<DatabaseTable
 		@"SELECT
                trim(I.rdb$index_name) as INDEX_NAME,
                COALESCE(I.rdb$unique_flag, 0) as IS_UNIQUE,
+               CASE Coalesce(I.rdb$index_type, 0)
+                   WHEN 0 THEN false
```
cincuranet, line 315: I guess this could be simplified.

cincuranet (changes requested):

src/FirebirdSql.EntityFrameworkCore.Firebird/Migrations/FbMigrationsSqlGenerator.cs
lweberprb: Good idea! Always better to be explicit.

lweberprb:
Sorry it took me so long to include your feedback in the PR. Please have a look at the changed files.

src/FirebirdSql.EntityFrameworkCore.Firebird/Migrations/FbMigrationsSqlGenerator.cs
```diff
@@ -201,6 +201,19 @@ protected override void Generate(CreateIndexOperation operation, IModel model, M
 		{
 			builder.Append("UNIQUE ");
 		}
+
+		if (operation.IsDescending is not null && operation.IsDescending.Length > 0)
+		{
+			var isDescending = operation.IsDescending[0];
+			if (operation.IsDescending.Any(x => x != isDescending))
+				throw new NotSupportedException("Mixed order indices are not supported by Firebird");
```
cincuranet, line 209: Please include comma at the end.

src/FirebirdSql.EntityFrameworkCore.Firebird/Scaffolding/Internal/FbDatabaseModelFactory.cs
```diff
@@ -343,11 +344,18 @@ private void GetIndexes(DbConnection connection, IReadOnlyList<DatabaseTable> ta
...
-						foreach (var column in reader.GetString(2).Split(','))
+						foreach (var column in reader.GetString(3).Split(','))
 						{
 							index.Columns.Add(table.Columns.Single(y => y.Name == column.Trim()));
 						}
 
+						if (reader.GetBoolean(2))
+						{
+							var isDescending = new bool[index.Columns.Count];
+							isDescending[0] = true;
```
cincuranet, line 355: Shouldn't this be set to `true` for all elements of the array?

cincuranet:
Few minor comments above. Otherwise LGTM.

src/FirebirdSql.EntityFrameworkCore.Firebird/Scaffolding/Internal/FbDatabaseModelFactory.cs
lweberprb: Correct, missed this while modifying the PR after your first review.

src/FirebirdSql.EntityFrameworkCore.Firebird/Migrations/FbMigrationsSqlGenerator.cs
lweberprb: Its included now

cincuranet (approved):

cincuranet:
Nice!

PR #122, opened 2023-07-27
RFC: Reduce memory allocations [Do not merge]

fdcastel:
Initial work to reduce memory allocations. 

Asking for guidance/opinions. Do not merge. 

### Before be56c23c7a623d7dfbe0132a2b6d479466d0ffc4
```
| Method | Count |             DataType |      Mean |     Error |   StdDev |       Gen0 |       Gen1 | Allocated |
|------- |------ |--------------------- |----------:|----------:|---------:|-----------:|-----------:|----------:|
|  Fetch | 10000 |               BIGINT |  56.14 ms |  0.714 ms | 0.039 ms |  1333.3333 |  1222.2222 |  10.86 MB |
|  Fetch | 10000 | BLOB SUB_TYPE BINARY | 182.51 ms | 56.430 ms | 3.093 ms | 24333.3333 | 12000.0000 | 196.11 MB |
|  Fetch | 10000 | BLOB (...) UTF8 [37] | 191.21 ms | 90.292 ms | 4.949 ms | 27000.0000 |  9000.0000 | 216.18 MB |
|  Fetch | 10000 | CHAR((...)CTETS [30] |  58.46 ms |  3.942 ms | 0.216 ms |  1625.0000 |   750.0000 |  13.31 MB |
|  Fetch | 10000 | CHAR((...) UTF8 [28] |  80.73 ms |  4.432 ms | 0.243 ms |  5666.6667 |  1333.3333 |  45.43 MB |
```

### After 6a641d7079db56c489c73dd3e5158b4319cee242
```
| Method | Count |             DataType |      Mean |     Error |   StdDev |       Gen0 |       Gen1 | Allocated |
|------- |------ |--------------------- |----------:|----------:|---------:|-----------:|-----------:|----------:|
|  Fetch | 10000 |               BIGINT |  22.34 ms |  4.823 ms | 0.264 ms |   593.7500 |   562.5000 |   4.91 MB |
|  Fetch | 10000 | BLOB SUB_TYPE BINARY | 139.91 ms | 59.043 ms | 3.236 ms | 23750.0000 | 11750.0000 | 190.15 MB |
|  Fetch | 10000 | BLOB (...) UTF8 [37] | 141.46 ms | 57.728 ms | 3.164 ms | 26250.0000 | 12750.0000 | 210.22 MB |
|  Fetch | 10000 | CHAR((...)CTETS [30] |  24.28 ms |  5.843 ms | 0.320 ms |   937.5000 |   906.2500 |   7.66 MB |
[...]

PR #130, opened 2024-10-25
Performance: Removes marshalling from NativeToManaged conversions.

fdcastel:
Another piece of #122.

- Reduce memory allocations in MarshalManagedToNative.
- Convert XSQLDA / XSQLVAR (internals) to structs.
- Appends _STRUCT suffix to XSQLDA / XSQLVAR to make the conversion path easier (can be removed, later).

### DIFF
--- a/src/FirebirdSql.Data.FirebirdClient/Client/Native/Marshalers/XSQLDA.cs
+++ b/src/FirebirdSql.Data.FirebirdClient/Client/Native/Marshalers/XSQLDA.cs
@@ -15,16 +15,12 @@
 
 //$Authors = Carlos Guzman Alvarez, Jiri Cincura (jiri@cincura.net)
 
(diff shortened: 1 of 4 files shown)


fdcastel:
_⚠️ I'm not seeing any changes in allocations when using .NET8. ⚠️_

It's worth noting that this PR was [initially created over a year ago](https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient/pull/122#issue-1825087075), and at that time, the allocation differences were more noticeable. 

One possible explanation could be that the .NET team has optimized allocation handling in Marshalling with the release of .NET 8.

cincuranet:
OK. I'm going to close this now @fdcastel. We can revisit it in the future if we see any room for improvement (for embedded there's a lot of potential).

PR #132, opened 2025-04-09
Add support for Int64 in DbValue.GetTime() when the value represents ticks of a TimeOnly property.

madaleno:
Some Orm frameworks like ServiceStack.OrmLite convert TimeOnly property values to Ticks and pass an Int64 value when creating parameters or set the field value, instead of passing a TimeOnly object. 
With this change it will be possible to use this kind of framework without getting conversion errors.


cincuranet:
This needs tests.

PR #133, opened 2025-04-21
Add missing error messages and refresh error messages

mrotteveel:
Source: Firebird 5.0.2.1613-0

Fixes: #1221


mrotteveel:
I notice that maybe I should have directed the PR to https://github.com/FirebirdSQL/NETProvider instead. Do you want me to correct that?

mrotteveel:
I included *all* error facilities, but if you want, I can excluded some (e.g. the ISQL facility is not really needed).

cincuranet:
Thanks.

